PR #3, opened 2023-01-09
Add job history model

yashmehrotra:


models/job_history.go
```diff
@@ -0,0 +1,72 @@
...
+	Name         string
+	SuccessCount int
+	ErrorCount   int
+	Hostname     string
+	TimeTakenMs  int64
+	ResourceType string
+	ResourceID   string
+	Details      types.JSONMap
+	Status       string
+	TimeStart    time.Time `gorm:"-"`
```
moshloop, line 27: I think we want to keep the start/end date-

models/job_history.go
```diff
@@ -0,0 +1,72 @@
...
+	StatusFinished = "FINISHED"
+)
+
+type JobHistory struct {
+	ID           uuid.UUID `gorm:"default:generate_ulid()"`
+	Name         string
+	SuccessCount int
+	ErrorCount   int
+	Hostname     string
+	TimeTakenMs  int64
```
moshloop, line 22:
```suggestion
	DurationMillis  int64
```

PR #8, opened 2023-01-11
feat: switch to atlas for db model

moshloop:


yashmehrotra:
@moshloop This is great 👍🏽

functions/functions.go
```diff
@@ -0,0 +1,24 @@
+package functions
+
+import "embed"
+
+//go:embed *.sql
+var functions embed.FS
+
+var Functions map[string]string
+
+func GetFunctions() (map[string]string, error) {
```
yashmehrotra, line 10: How is this being called ?

yashmehrotra (approved):

yashmehrotra:
@moshloop For triggers, we might have to do a `DROP IF EXISTS ... CREATE TRIGGER` in a always kind of a script

functions/functions.go
moshloop: https://github.com/flanksource/incident-commander/pull/218/files#diff-953a885e28970388a5ee8146da2db0488f6c2539fe2db2caca45a4033f090985R138

PR #32, opened 2023-02-16
Struct migrations from canary checker

adityathebe:


models/checks.go
```diff
@@ -0,0 +1,849 @@
...
+import (
+	"encoding/json"
+	"fmt"
+	"strings"
+
+	"github.com/flanksource/kommons"
+	v1 "k8s.io/api/core/v1"
+)
+
+type ICheck interface {
```
adityathebe, line 12: @yashmehrotra had to rename this interface from `Check`. It was in `canary-checker/pkg/api/external`

models/components.go
```diff
@@ -0,0 +1,539 @@
...
+	CreatedAt        time.Time  `gorm:"column:created_at" json:"created_at,omitempty"`
+	UpdatedAt        time.Time  `gorm:"column:updated_at" json:"updated_at,omitempty"`
+	DeletedAt        *time.Time `gorm:"column:deleted_at" json:"deleted_at,omitempty"`
+}
+
+func (cr ComponentRelationship) TableName() string {
+	return "component_relationships"
+}
+
+type ConfigComponentRelationship struct {
```
adityathebe, line 528: This came from `canary-checker/pkg/db`. Added the `gorm` tags & the last 3 fields.

adityathebe:
Most of what's in here is just copy-paste. For the few things where I had to intervene, I've left some comments.

models/canary.go
```diff
@@ -0,0 +1,274 @@
...
+// 		Source:    canary.Annotations["source"],
+// 		Schedule:  canary.Spec.GetSchedule(),
+// 		Checks:    types.JSONStringMap(checks),
+// 	}, nil
+// }
+
+type ResultMode string
+
+// CanarySpec defines the desired state of Canary
+type CanarySpec struct {
```
yashmehrotra, line 81: This can be removed

yashmehrotra: Everything below this can be

models/checks.go
```diff
@@ -0,0 +1,849 @@
+package models
```
yashmehrotra, line 1: I think this entire file can be removed

models/components.go
```diff
@@ -0,0 +1,539 @@
...
+	Configs          Configs                  `json:"configs,omitempty" gorm:"type:configs"`
+	CostPerMinute    float64                  `json:"cost_per_minute,omitempty" gorm:"column:cost_per_minute"`
+	CostTotal1d      float64                  `json:"cost_total_1d,omitempty" gorm:"column:cost_total_1d"`
+	CostTotal30d     float64                  `json:"cost_total_30d,omitempty" gorm:"column:cost_total_30d"`
+	CostTotal7d      float64                  `json:"cost_total_7d,omitempty" gorm:"column:cost_total_7d"`
+	CreatedAt        time.Time                `json:"created_at,omitempty" time_format:"postgres_timestamp"`
+	DeletedAt        *time.Time               `json:"deleted_at,omitempty" time_format:"postgres_timestamp" swaggerignore:"true"`
+	ExternalId       string                   `json:"external_id,omitempty"` //nolint
+	Icon             string                   `json:"icon,omitempty"`
+	ID               uuid.UUID                `json:"id,omitempty" gorm:"default:generate_ulid()"` //nolint
```
yashmehrotra, line 36: Keep ID as the first, looks better

models/components.go
yashmehrotra: Gorm tags might not be required since the tags we added are the default anyway

models/db_types.go
```diff
@@ -0,0 +1,11 @@
+package models
```
yashmehrotra, line 1: Move this to `types/types.go`

models/varsource.go
```diff
@@ -0,0 +1,164 @@
...
+	"k8s.io/apimachinery/pkg/api/meta"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	runtime "k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/apimachinery/pkg/util/sets"
+	"k8s.io/apimachinery/pkg/util/validation"
+	"k8s.io/client-go/kubernetes"
+)
+
+// VarSource represents a source for a value
+type VarSource struct {
```
yashmehrotra, line 18: This is a k8s entity, might not be required

yashmehrotra (changes requested):
Might have to delete more things from here

models/components.go
```diff
@@ -0,0 +1,539 @@
...
+	"gorm.io/gorm"
+	"gorm.io/gorm/clause"
+	"gorm.io/gorm/schema"
+)
+
+type Components []*Component
+
+type Component struct {
+	Checks           Checks                   `json:"checks,omitempty" gorm:"-"`
+	ComponentChecks  ComponentChecks          `json:"-" gorm:"column:component_checks" swaggerignore:"true"`
```
yashmehrotra, line 24: @moshloop Do we have a need for this field ?

models/config.go
```diff
@@ -1,16 +1,79 @@
...
 
 	"github.com/flanksource/duty/types"
+	"github.com/google/uuid"
 	"github.com/lib/pq"
 	"gorm.io/gorm"
+	"gorm.io/gorm/clause"
+	"gorm.io/gorm/schema"
 )
 
+type Config struct {
```
yashmehrotra, line 20: Use ConfigItem instead

models/components.go
adityathebe: okay. I generally sort fields in a relatively large struct so it's easier to scan

models/config.go
adityathebe: @yashmehrotra Does this mean `models.Config` can be deleted?

models/checks.go
adityathebe: It's removed

models/config.go
yashmehrotra: Yeah, they refer to the same thing

types/types.go
```diff
@@ -17,6 +17,17 @@ import (
...
+	SQLServerType = "sqlserver"
+	PostgresType  = "postgres"
+	SqliteType    = "sqlite"
+	MysqlType     = "mysql"
+	Text          = "TEXT"
+	JSONType      = "JSON"
+	JSONBType     = "JSONB"
+	NVarcharType  = "NVARCHAR(MAX)"
+)
+
```
yashmehrotra, line 30: 👍🏽

models/components.go
```diff
@@ -0,0 +1,482 @@
...
+	Unit string `json:"unit,omitempty"`
+	Max  *int64 `json:"max,omitempty"`
+	Min  int64  `json:"min,omitempty"`
+
+	Status         string `json:"status,omitempty"`
+	LastTransition string `json:"lastTransition,omitempty"`
+	Links          []Link `json:"links,omitempty"`
+}
+
+func (p Property) GetValue() interface{} {
```
yashmehrotra, line 180: prefer `any` instead of `interface{}`

go.mod
```diff
@@ -4,8 +4,8 @@ go 1.19
 
 require (
 	ariga.io/atlas v0.9.0
+	github.com/fergusstrange/embedded-postgres v1.19.0
```
yashmehrotra, line 7: which package is requiring this ?

models/components.go
```diff
@@ -0,0 +1,482 @@
...
+	}
+	return ""
+}
+
+func (p Properties) GormValue(ctx context.Context, db *gorm.DB) clause.Expr {
+	data, _ := json.Marshal(p)
+	return gorm.Expr("?", data)
+}
+
+type Check struct {
```
yashmehrotra, line 330: can move this to models/checks.go

go.mod
adityathebe: `schema_test.go`

models/config.go
adityathebe: okay, deleted it.

yashmehrotra (approved):

PR #39, opened 2023-02-21
feat: add generic gorm scan/value funcs

yashmehrotra:
With this we can avoid a lot of boilerplate code when working with json based structs with gorm. We wouldn't have to write Scan/Value() funcs everytime.

```go
type Person struct {
	ID         uuid.UUID        `json:"id" gorm:"default:generate_ulid()"`
	Name       string           `json:"name,omitempty"`
	Properties PersonProperties `json:"properties,omitempty"`
}

type PersonProperties struct {
	Car   string `json:"car,omitempty"`
	Wheel string `json:"wheel,omitempty"`
}

func (p PersonProperties) Value() (driver.Value, error) {
	return types.GenericStructValue(p, true)
}

func (p *PersonProperties) Scan(val any) error {
	return types.GenericStructScan(&p, val)
}
```

### DIFF
--- a/types/types.go
+++ b/types/types.go
@@ -355,3 +355,34 @@ func (jm JSONMap) GormValue(ctx context.Context, db *gorm.DB) clause.Expr {
 	data, _ := jm.MarshalJSON()
 	return gorm.Expr("?", string(data))
 }
+
+// GenericStructValue can be set as the Value() func for any json struct
+func GenericStructValue[T any](t T, defaultNull bool) (driver.Value, error) {
+	b, err := json.Marshal(t)
+	if err != nil {
+		return b, err
+	}
+	if defaultNull && string(b) == "{}" {
+		return nil, nil
+	}
+	return string(b), nil
+}
+
+// GenericStructScan can be set as the Scan(val) func for any json struct
+func GenericStructScan[T any](t *T, val any) error {
+	if val == nil {
+		*t = *new(T)
+		return nil
+	}
+	var ba []byte
+	switch v := val.(type) {
+	case []byte:
+		ba = v
+	case string:
+		ba = []byte(v)
+	default:
+		return errors.New(fmt.Sprint("Failed to unmarshal JSONB value:", val))
+	}
+	err := json.Unmarshal(ba, &t)
+	return err
+}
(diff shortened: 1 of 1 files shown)


types/types.go
```diff
@@ -355,3 +355,34 @@ func (jm JSONMap) GormValue(ctx context.Context, db *gorm.DB) clause.Expr {
 	data, _ := jm.MarshalJSON()
 	return gorm.Expr("?", string(data))
 }
+
+// GenericStructValue can be set as the Value() func for any json struct
+func GenericStructValue[T any](t T, defaultNull bool) (driver.Value, error) {
```
yashmehrotra, line 360: Is this name okay or should I rename it to `GenericStructJSONValue` ?

types/types.go
```diff
@@ -355,3 +355,34 @@ func (jm JSONMap) GormValue(ctx context.Context, db *gorm.DB) clause.Expr {
...
+	if defaultNull && string(b) == "{}" {
+		return nil, nil
+	}
+	return string(b), nil
+}
+
+// GenericStructScan can be set as the Scan(val) func for any json struct
+func GenericStructScan[T any](t *T, val any) error {
+	if val == nil {
+		*t = *new(T)
```
adityathebe, line 374: If `t` is a nil pointer then this panics. But the way we're using this func guarantees what it won't ever be nil. So I guess it's fine.

Although this could just be `t = new(T)`.

types/types.go
```diff
@@ -355,3 +355,34 @@ func (jm JSONMap) GormValue(ctx context.Context, db *gorm.DB) clause.Expr {
...
+		return nil
+	}
+	var ba []byte
+	switch v := val.(type) {
+	case []byte:
+		ba = v
+	case string:
+		ba = []byte(v)
+	default:
+		return errors.New(fmt.Sprint("Failed to unmarshal JSONB value:", val))
```
adityathebe, line 384: I personally prefer

`fmt.Errorf("failed to unmarshal JSONB value: %v", val)`

types/types.go
adityathebe: fine with both 👍

types/types.go
yashmehrotra: 👍🏽

types/types.go
yashmehrotra: 👍🏽

PR #49, opened 2023-02-28
chore: topology api refactor

yashmehrotra:


models/components.go
```diff
@@ -133,6 +131,20 @@ func (c *Component) Summarize() Summary {
 	return s
 }
 
+type Components []*Component
+
+func (components Components) Walk() Components {
```
moshloop, line 136: This is going to create many new objects - Maybe we should change this to return a channel?

yashmehrotra: Update: Rename to map and pass a component modifying func

```
func (components Components) Map(func (*models.Component ) ) Components {
```

yashmehrotra: 👍🏽

PR #50, opened 2023-02-28
Added a trigger to all tables that have "updated_at" column.

adityathebe:
On row updates, the `updated_at` column value is set to `Now()`.

Closes https://github.com/flanksource/duty/issues/48

### DIFF
--- a/views/013_updated_at_column_trigger.sql
+++ b/views/013_updated_at_column_trigger.sql
@@ -1,0 +1,41 @@
+CREATE EXTENSION IF NOT EXISTS hstore;
+
+CREATE OR REPLACE FUNCTION update_updated_at_column()
+RETURNS TRIGGER AS $$
+DECLARE
+  changed_fields hstore;
+BEGIN
+  IF NOT (TG_WHEN = 'BEFORE' AND TG_OP = 'UPDATE') THEN
+    RAISE EXCEPTION 'update_updated_at_column() should only run as a BEFORE UPDATE trigger';
+  END IF;
+
+  changed_fields = hstore(NEW.*) - hstore(OLD.*);
+  IF changed_fields = hstore('') THEN
+    RETURN NULL; -- No columns have been updated.
+  END IF;
+
+  NEW.updated_at = NOW();
+  RETURN NEW;
+END;
+$$ LANGUAGE plpgsql;
+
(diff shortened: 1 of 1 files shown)


views/013_updated_at_column_trigger.sql
```diff
@@ -0,0 +1,28 @@
...
+BEGIN
+  FOR tbl_name IN
+    SELECT table_name
+    FROM information_schema.columns
+    WHERE 
+      table_schema = current_schema() 
+      AND column_name = 'updated_at'
+      AND table_name NOT IN (SELECT table_name FROM information_schema.views WHERE table_schema = current_schema()) 
+  LOOP
+    EXECUTE format('CREATE OR REPLACE TRIGGER %I_update_updated_at
```
moshloop, line 23: Does this not loop? Do we not need fire if a) any of the fields have changed and b) updated_at is not in the change list ?

adityathebe: No, this does not run into an infinite loop because the `update_updated_at_column` trigger function simply modifies the row before update happens. The function itself doesn't run any SQL statements to perform the update.

> any of the fields have changed

I've made this modification so the `updated_at` column is set to `NOW()` only if there's some modification to the row that's being updated.

views/013_updated_at_column_trigger.sql
```diff
@@ -0,0 +1,41 @@
...
+CREATE OR REPLACE FUNCTION update_updated_at_column()
+RETURNS TRIGGER AS $$
+DECLARE
+  changed_fields hstore;
+BEGIN
+  IF NOT (TG_WHEN = 'BEFORE' AND TG_OP = 'UPDATE') THEN
+    RAISE EXCEPTION 'update_updated_at_column() should only run as a BEFORE UPDATE trigger';
+  END IF;
+
+  changed_fields = hstore(NEW.*) - hstore(OLD.*);
```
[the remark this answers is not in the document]
adityathebe: Got this from https://github.com/2ndQuadrant/audit-trigger/blob/master/audit.sql#L125-L130

moshloop:
@adityathebe can you rebase ?

adityathebe:
> @adityathebe can you rebase ?

@moshloop Done 👍

PR #54, opened 2023-03-03
feat: add transition time for checks

yashmehrotra:


views/013_check_summary.sql
```diff
@@ -0,0 +1,31 @@
...
+        check_statuses
+    WHERE
+        time > (NOW() at TIME ZONE 'utc' - Interval '1 hour')
+    GROUP BY check_id;
+
+-- For last transition
+CREATE OR REPLACE FUNCTION update_last_transition_time_for_check()
+RETURNS TRIGGER AS $$
+BEGIN
+    UPDATE checks SET last_transition_time = NOW() WHERE id = NEW.id;
```
moshloop, line 21: shouldn't this be NEW.last_transition_time = now() with a before trigger ?

yashmehrotra: yeah, that is much more efficient

PR #68, opened 2023-03-09
feat: aws view for config_items

yashmehrotra:


moshloop:
I think we should use the existing configs view as base and then add ontop of that

moshloop:
@yashmehrotra  https://github.com/flanksource/duty/blob/main/views/006_config_views.sql#L3 has more than just the default columns

moshloop:
So maybe we change the select to be configs rather than config_items ?

yashmehrotra:
@moshloop Makes sense, have updated the query

PR #75, opened 2023-03-22
feat: add logging_backend model and schema

yashmehrotra:


schema/logging_backends.hcl
```diff
@@ -0,0 +1,48 @@
...
+  }
+  column "source" {
+    null = true
+    type = enum.source
+  }
+  column "created_at" {
+    null = true
+    type = timestamptz
+  }
+  column "created_by" {
```
moshloop, line 33: I think we need foreign key

schema/logging_backends.hcl
```diff
@@ -0,0 +1,48 @@
...
+    null = false
+    type = jsonb
+  }
+  column "source" {
+    null = true
+    type = enum.source
+  }
+  column "created_at" {
+    null = true
+    type = timestamptz
```
moshloop, line 31: I think this should be timestamp, when we switch to timestampz it should be for everything

yashmehrotra: I was thinking the reverse, that lets use timestamptz for new tables, and do a big bang for everything later

PR #76, opened 2023-03-22
Agents and connections

moshloop:
Agents are meant to represent push clients - we might need to update some of the constraints lookups to use it
Connections are for re-using credentials without needing access / exposing them


adityathebe (approved):
LGTM

PR #80, opened 2023-03-27
feat: added Query func to run generic SQL queries

adityathebe:


query.go
```diff
@@ -0,0 +1,66 @@
...
+	"database/sql"
+	"fmt"
+
+	"github.com/flanksource/commons/logger"
+)
+
+// Query runs the given SQL query against the provided db connection.
+// The rows are returned as a map of columnName=>columnValue.
+func Query(ctx context.Context, conn *sql.DB, query string) ([]map[string]any, error) {
+	tx, err := conn.BeginTx(ctx, &sql.TxOptions{ReadOnly: true})
```
moshloop, line 14: I think use dbpool *pgxpool.Pool, to make it consistent

moshloop, line 14: I don't think we need to start and commit a transaction for a readonly query

adityathebe: That's just to ensure that only read queries are run.

query.go
```diff
@@ -0,0 +1,53 @@
...
+
+		results = append(results, row)
+	}
+
+	err = rows.Err()
+	if err != nil {
+		return nil, err
+	}
+
+	return results, nil
```
moshloop, line 52: nit: You could do

```golang

if err := rows.Err(); err != nil {
return nil, err
}
```

or even just `return results, rows.Err()`

PR #101, opened 2023-04-05
feat: Json diff

adityathebe:
removes patch column in config_changes table.

### DIFF
--- a/models/config.go
+++ b/models/config.go
@@ -79,7 +79,7 @@ type ConfigChange struct {
 	Severity         string     `gorm:"column:severity" json:"severity"  faker:"oneof: critical, high, medium, low, info"`
 	Source           string     `gorm:"column:source" json:"source"`
 	Summary          string     `gorm:"column:summary;default:null" json:"summary,omitempty"`
-	Patches          string     `gorm:"column:patches;default:null" json:"patches,omitempty"`
+	Diff             string     `gorm:"column:diff;default:null" json:"diff,omitempty"`
 	Details          types.JSON `gorm:"column:details" json:"details,omitempty"`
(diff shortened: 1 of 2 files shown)


models/config.go
```diff
@@ -79,7 +80,7 @@ type ConfigChange struct {
 	Severity         string     `gorm:"column:severity" json:"severity"  faker:"oneof: critical, high, medium, low, info"`
 	Source           string     `gorm:"column:source" json:"source"`
 	Summary          string     `gorm:"column:summary;default:null" json:"summary,omitempty"`
-	Patches          string     `gorm:"column:patches;default:null" json:"patches,omitempty"`
```
moshloop, line 82: I don't think we want to remove the patch - This would still be useful for querying specific changes - the diff is in addition

models/config.go
```diff
@@ -99,6 +101,23 @@ func (c ConfigChange) String() string {
...
 
+// BeforeCreate is a user defined hook for Gorm.
+// It will be called when creating a record.
+func (c *ConfigChange) BeforeCreate(tx *gorm.DB) (err error) {
+	if c.ID == "" {
+		c.ID = uuid.New().String()
+	}
+
+	if c.CreatedAt == nil {
+		now := time.Now()
```
moshloop, line 112: I don't think we can use server time  - it needs to use the db time = it should do it already with the db default value ?

adityathebe: @moshloop that's correct - should be populated in DB. Not sure why it was like that before. I just copy-pasted this `BeforeCreate` hook from config db models as I'm getting rid of the db models from there

https://github.com/flanksource/config-db/blob/main/db/models/config_change.go#L63-L74

adityathebe: Made the change

PR #103, opened 2023-04-07
move log selector to types

adityathebe:


moshloop:
@adityathebe nothing in the .bin file should be committed, Can you rollback ?

adityathebe:
@moshloop sorry didn't notice that

Issue #105, opened 2023-04-10
Incident summary view


moshloop:
i.e. Just title, severity, type, status, created, last_updated, distinct responders / commenters / commander

adityathebe:
@moshloop I'm not sure what `commander` means here.

PR #106, opened 2023-04-10
feat: added incident_summary view

adityathebe:
Resolves https://github.com/flanksource/duty/issues/105


adityathebe:
```
-[ RECORD 1 ]-------------------------
id                  | 8305385a-0e76-490b-a5d2-e51c79bcfe25
title               | Broken
severity            | Low
type                | availability
status              | open
created_at          | 2023-03-21 12:09:37.193753
updated_at          | 2023-03-21 12:09:37.193753
commander           | {"id": "aab5b129-4195-4a65-b0d3-e0adf109d848", "name": "Aditya Thebe", "avatar": null}
distinct_responders | [{"id": "aab5b129-4195-4a65-b0d3-e0adf109d848", "name": "Aditya Thebe", "avatar": null}]
distinct_commenters | <null>
```

moshloop:
@adityathebe I think just call them responders and commenters

moshloop:
@adityathebe can we add a test for this query - if we can add dummy data quickly then we can verify the results but at a minimum we should verify that the views runs

fixtures/dummy/incidents.go
```diff
@@ -24,32 +24,51 @@ var FirstComment = models.Comment{
...
 	UpdatedAt:  time.Now(),
 }
 
-var AllDummyComments = []models.Comment{FirstComment, SecondComment}
+var AllDummyComments = []models.Comment{FirstComment, SecondComment, ThirdComment}
 
 var FirstResponder = models.Responder{
 	ID:         uuid.New(),
 	IncidentID: LogisticsAPIDownIncident.ID,
 	Type:       "whattype",
```
yashmehrotra, line 54: Can you change this to a Generic Type like `Jira` or `MSPlanner` ?

fixtures/dummy/incidents.go
```diff
@@ -24,32 +24,51 @@ var FirstComment = models.Comment{
 	CreatedBy:  JohnWick.ID,
 	Comment:    "This is a comment",
 	IncidentID: LogisticsAPIDownIncident.ID,
-	CreatedAt:  time.Now(),
-	UpdatedAt:  time.Now(),
+	CreatedAt:  time.Now().Add(-time.Hour),
```
yashmehrotra, line 27: Why are we doing this ?

adityathebe: The test fails sometimes and I think it's because of the unstable ordering of the comments, Although I am not able to reproduce it enough times to make sure of it. I tried a few different ways to rewrite the SQL query to order the comment but the query was too verbose.

I was just experimenting with time to see if it helps in stable ordering.

adityathebe: I have modified the query

PR #112, opened 2023-04-13
chore: create index for canary_id, transformed in checks

yashmehrotra:


yashmehrotra:
FYI
@moshloop Forgot to add the index

PR #114, opened 2023-04-13
modify topology query to return extra metadata

adityathebe:


topology.go
```diff
@@ -156,7 +164,51 @@ func QueryTopology(dbpool *pgxpool.Pool, params TopologyOptions) ([]*models.Comp
...
+
+	placeholders := make([]string, len(componentIDs))
+	for i := range componentIDs {
+		placeholders[i] = fmt.Sprintf("$%d", i+1)
+	}
+
+	args := make([]any, len(componentIDs))
+	for i, id := range componentIDs {
+		args[i] = id
+	}
```
adityathebe, line 192: would much prefer to use gorm instead of doing this primitive stuff here but the `QueryTopology` func is implemented to work with `pgxpool.Pool`. 

Let me know if change it so that it accepts `gorm.DB` instead.

topology.go
```diff
@@ -156,7 +164,51 @@ func QueryTopology(dbpool *pgxpool.Pool, params TopologyOptions) ([]*models.Comp
...
+	res.Teams, err = GetTeamNamesOfComponents(ctx, dbpool, res.componentIDs)
+	if err != nil {
+		return nil, fmt.Errorf("failed to get teams of components: %w", err)
+	}
+
+	return &res, nil
+}
+
+func GetTeamNamesOfComponents(ctx context.Context, dbpool *pgxpool.Pool, componentIDs []string) ([]string, error) {
+	if len(componentIDs) == 0 {
```
moshloop, line 180: Can we not just update the views to return teams per component?

topology.go
```diff
@@ -156,7 +164,51 @@ func QueryTopology(dbpool *pgxpool.Pool, params TopologyOptions) ([]*models.Comp
...
+	for i := range componentIDs {
+		placeholders[i] = fmt.Sprintf("$%d", i+1)
+	}
+
+	args := make([]any, len(componentIDs))
+	for i, id := range componentIDs {
+		args[i] = id
+	}
+
+	query := fmt.Sprintf("SELECT DISTINCT teams.name FROM team_components LEFT JOIN teams ON teams.id = team_components.team_id WHERE component_id IN (%s)", strings.Join(placeholders, ","))
```
yashmehrotra, line 194: Try to make this part of the Topology Query, not a separate query

topology.go
```diff
@@ -290,3 +342,69 @@ func GetComponent(ctx context.Context, db *gorm.DB, id string) (*models.Componen
...
+		if _, exists := t.tagMap[tagKey]; !exists {
+			t.Tags = append(t.Tags, Tag{Key: k, Val: v})
+			t.tagMap[tagKey] = struct{}{}
+		}
+	}
+}
+
+// populateTopologyResult goes through the components recursively (depth-first)
+// and populates the TopologyResponse struct.
+func populateTopologyResult(components models.Components, res *TopologyResponse) {
```
yashmehrotra, line 402: s/TopologyResult/TopologyResponse

topology.go
```diff
@@ -290,3 +342,69 @@ func GetComponent(ctx context.Context, db *gorm.DB, id string) (*models.Componen
...
+	for _, component := range components {
+		res.componentIDs = append(res.componentIDs, component.ID.String())
+		res.AddTag(component.Labels)
+		res.AddType(component.Type)
+		res.AddHealthStatuses(string(component.Status))
+		populateTopologyResult(component.Components, res)
+	}
+}
```
yashmehrotra, line 410: Can we do this at the tree creation level ? We are already iterating through all the components there once, we shouldn't do it twice

topology.go
```diff
@@ -290,3 +342,69 @@ func GetComponent(ctx context.Context, db *gorm.DB, id string) (*models.Componen
...
+	HealthStatuses []string          `json:"healthStatuses,omitempty"`
+	Teams          []string          `json:"teams,omitempty"`
+	Tags           []Tag             `json:"tags,omitempty"`
+	Types          []string          `json:"types,omitempty"`
+}
+
+func (t *TopologyResponse) AddHealthStatuses(s string) {
+	if t.healthStatusMap == nil {
+		t.healthStatusMap = make(map[string]struct{})
+	}
```
yashmehrotra, line 367: You can create a `NewTopologyResponse` func which returns a TopologyResponse with components and maps already initialized

adityathebe: I chose not to do that since it's very easy to directly instantiate the variable without using a constructor.

yashmehrotra: I think if you use a CTE, you wouldn't have to do it in code, it will be handled by DB instead using `Distinct`

Something like,
```
WITH topology AS (
  <original_query>
)

SELECT json_build_object('components', to_json(topology), 'tags', SELECT DISTINCT(tags) FROM topology, ... and so on) FROM topology;
```

topology.go
adityathebe: Leavining this as it is for now since we'll be moving to topology view.

topology.go
adityathebe: We're not using the `topology` view right now. I'll leave this as it is for now since we'll be migrating to views anyway.

topology.go
adityathebe: Tree creation doesn't always happen (_when you need flat structure_) so it's not appropriate to do this at that level. And we're already iterating through components numerous times with all the filters.

Regardless, this might not even be needed once we move to view. So, I'm leaving this as it is.

yashmehrotra: Yeah, doing it in tree creation is not a good idea

I don't think populateTopology function is required anyway, since the query will use `SELECT DISTINCT` for all the unique sets we need

topology.go
yashmehrotra: @adityathebe its better to remove this as discussed in yesterday's call

Fixing the topology view should be the priority

yashmehrotra (changes requested):

topology.go
adityathebe: I must have misheard but I thought that wasn't the priority right now - adding metadata was.

yashmehrotra: Yeah, basically, skip teams for now, add other metadata and then do topology view with teams

topology.go
```diff
@@ -290,3 +342,69 @@ func GetComponent(ctx context.Context, db *gorm.DB, id string) (*models.Componen
...
+		if _, exists := t.tagMap[tagKey]; !exists {
+			t.Tags = append(t.Tags, Tag{Key: k, Val: v})
+			t.tagMap[tagKey] = struct{}{}
+		}
+	}
+}
+
+// populateTopologyResponse goes through the components recursively (depth-first)
+// and populates the TopologyResponse struct.
+func populateTopologyResponse(components models.Components, res *TopologyResponse) {
```
moshloop, line 402:
```suggestion
func addMetadata(components models.Components, res *TopologyResponse) {
```

moshloop (approved):

yashmehrotra (approved):

Issue #117, opened 2023-04-13
Switch to new topology view


moshloop:

adityathebe:
Related to https://github.com/flanksource/duty/issues/97

yashmehrotra:
Addressed by https://github.com/flanksource/duty/pull/129

PR #124, opened 2023-04-18
feat: remove old query for topology. Use existing view.

adityathebe:
Resolves https://github.com/flanksource/duty/issues/117

### DIFF
--- a/topology.go
+++ b/topology.go
@@ -59,20 +59,25 @@ SELECT
 FROM
   topology`
 
+	var clauses []string
 	args := make(map[string]any)
 	if opts.ID != "" {
-		query = fmt.Sprintf("%s WHERE id = @id OR parent_id = @id", query)
+		clauses = append(clauses, "(id = @id OR parent_id = @id)")
 		args["id"] = opts.ID
 	}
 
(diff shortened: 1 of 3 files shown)


topology.go
```diff
@@ -24,123 +23,122 @@ type TopologyOptions struct {
...
+		&c.Path,
+		&c.Summary,
+		&c.IsLeaf,
+		&c.CreatedBy,
+		&c.CreatedAt,
+		&c.UpdatedAt,
+		&c.DeletedAt,
+		&c.LogSelectors,
+		&c.Checks,
+	)
```
yashmehrotra, line 119: https://pkg.go.dev/github.com/jackc/pgx/v5#RowToStructByPos

Prefer this or a gorm struct

If there are a lot of columns with index being important, it may lead to off by one or other indexing mistakes

adityathebe: I'm not really seeing the benefit with `RowToStructByPos`. It looks like you still need to maintain the order in the newly defined struct, no?

adityathebe:
Closing this. Addressed by https://github.com/flanksource/duty/pull/129

PR #127, opened 2023-04-20
added incident_id column

adityathebe:
https://github.com/flanksource/incident-commander/issues/123

### DIFF
--- a/functions/incident_ids.sql
+++ b/functions/incident_ids.sql
@@ -1,0 +1,13 @@
+CREATE SEQUENCE IF NOT EXISTS incident_id_sequence INCREMENT 1 START 1;
+
+CREATE OR REPLACE FUNCTION format_incident_id(seq_value BIGINT) 
+RETURNS VARCHAR
+AS $$
+  DECLARE
+    result VARCHAR;
+  BEGIN
+    result := 'INC' || LPAD(seq_value::TEXT, 7, '0');
+    RETURN result;
+  END;
+$$ 
+LANGUAGE plpgsql;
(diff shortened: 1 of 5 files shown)


functions/incident_ids.sql
```diff
@@ -0,0 +1,13 @@
+CREATE SEQUENCE IF NOT EXISTS incident_id_sequence INCREMENT 1 START 1;
```
yashmehrotra, line 1: @moshloop  @adityathebe 

Should we keep this in `functions`  or `views`

By name, `functions` make more sense, but we already have many functions in the `views` folder

In my mind, the distinction was that generic things go to functions and business logic related things go to `views`

Also, there are tons of functions/triggers in `views`. Should we eventually move them to `functions`  then?

Let me know your thoughts

schema/incidents.hcl
```diff
@@ -396,13 +395,19 @@ table "incident_rules" {
...
   column "id" {
     null    = false
     type    = uuid
     default = sql("generate_ulid()")
   }
+  column "incident_id" {
+    null = false
+    type = varchar(10)
+    default = sql("format_incident_id(NEXTVAL('incident_id_sequence'))")
+  }
```
yashmehrotra, line 410: This should have a unique clause too

functions/incident_ids.sql
adityathebe: This one in particular cannot be in `views` as the table column refers to it.

schema/incidents.hcl
adityathebe: I've added it but I initially didn't do it because we're already using database sequence. But it does make sense to have it for absolute safety 👍

moshloop:
@adityathebe I think we need add it into the incident summary

adityathebe:
@moshloop Added it to the `incident_summary` view and also added test for it.

PR #129, opened 2023-04-21
Use topology view in query

yashmehrotra:
@moshloop @adityathebe  You can ignore the *.sql changes, these are just fmt'd


fixtures/expectations/topology_depth_1_child_tree.json
```diff
@@ -1,33 +1,53 @@
...
+    "healthy",
+    "unhealthy"
+  ],
+  "tags": [
+    {
+      "telemetry": [
+        "enabled"
+      ]
+    }
+  ],
```
adityathebe, line 12: maybe tags could just be an object instead of an array with objects, if order doesn't matter here. Not sure how this will be structured if there were multiple tags, but I think this wouldn't be an ideal way to access the tags from the frontend.

So, instead of 
```json
{
  "tags": [
    {
      "telemetry": ["enabled"]
    },
    {
      "security": ["critical", "high"]
    }
  ]
}
```

we could do this
```json
{
  "tags": 
    {
      "telemetry": ["enabled"],
      "security": ["critical", "high"]
    }
}
```

fixtures/expectations/topology_depth_1_child_tree.json
```diff
@@ -1,33 +1,53 @@
...
+  "teams": [
+    ""
+  ],
+  "types": [
+    "Application",
+    "Database"
+  ],
```
adityathebe, line 19: Shouldn't this only contain "Application" type?

topology.go
```diff
@@ -20,143 +20,134 @@ type TopologyOptions struct {
...
 	for rows.Next() {
-		var components models.Components
 		if rows.RawValues()[0] == nil {
 			continue
 		}
 
-		if err := json.Unmarshal(rows.RawValues()[0], &components); err != nil {
-			return nil, fmt.Errorf("failed to unmarshal components:%v for %s", err, rows.RawValues()[0])
+		if err := json.Unmarshal(rows.RawValues()[0], &response); err != nil {
+			return nil, fmt.Errorf("failed to unmarshal TopologyResponse:%v for %s", err, rows.RawValues()[0])
```
adityathebe, line 132: This should be `%w` instead of `%v`

fixtures/expectations/topology_child_tree.json
```diff
@@ -1,51 +1,61 @@
...
-        "is_leaf": false,
-        "created_at": "2022-12-31T23:59:00",
-        "updated_at": "2022-12-31T23:59:00",
-        "relationship_id": "018681fe-f5aa-37e9-83f7-47b5b0232d5e"
+{
+  "healthStatuses": [
+    "healthy"
+  ],
+  "teams": [
+    ""
```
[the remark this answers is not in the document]
yashmehrotra: 👍🏽 

Have fixed it

fixtures/expectations/topology_depth_1_child_tree.json
yashmehrotra: Yeah, makes sense

👍🏽

Issue #77, opened 2023-03-23
Config query validation


moshloop:
Ensure sql is only selecting from configs_* tables

PR #138, opened 2023-05-01
feat: config query validation

adityathebe:
Related to https://github.com/flanksource/duty/issues/77


query.go
```diff
@@ -3,11 +3,105 @@ package duty
 import (
 	"context"
 	"fmt"
+	"strings"
```
moshloop, line 7: Can we move this into a `config` package ?

moshloop: @adityathebe ^

adityathebe:
Need to refactor. Ginkgo test suites inside `config` package isn't running.

config/query_test.go
```diff
@@ -0,0 +1,113 @@
...
+			t.Fail()
+		}
+		if output != test.expectedOutput {
+			t.Logf("%s: validateTablesInQuery() returned %v, but expected %v", test.name, output, test.expectedOutput)
+			t.Fail()
+		}
+	}
+}
+
+// var _ = ginkgo.Describe("Query should only run non mutation queries", func() {
```
adityathebe, line 84: This ginkgo spec is not moved to pkg_config_test.go because we're testing an unexported function.

PR #139, opened 2023-05-02
prepare for duty migration in canary checker

adityathebe:
Add new methods and some refactoring to help with https://github.com/flanksource/canary-checker/issues/794


models/components.go
```diff
@@ -76,17 +67,17 @@ type Component struct {
 	Parents        []string                  `json:"parents,omitempty" gorm:"-"`
 }
 
-func (c *Component) GetStatus() ComponentStatus {
+func (c *Component) GetStatus() types.ComponentStatus {
```
adityathebe, line 70: Though I would just use GetStatus() method from Summary

```go
func (c *Component) GetStatus() types.ComponentStatus {
	return c.Summary.GetStatus()
}
```
but they have different logics. Not sure if it's correct to do it this way.

adityathebe: Here, I see that component's status is being set using Summary's `GetStatus` method

https://github.com/flanksource/canary-checker/blob/master/pkg/topology/run.go#L399

types/component.go
```diff
@@ -0,0 +1,186 @@
...
+	processed bool
+}
+
+func (s *Summary) SetProcessed(val bool) {
+	s.processed = val
+}
+
+func (s Summary) IsProcessed() bool {
+	return s.processed
+}
```
yashmehrotra, line 43: These should be private (lowerCase)

types/component.go
```diff
@@ -0,0 +1,186 @@
...
+
+// GormDBDataType gorm db data type
+func (ResourceSelectors) GormDBDataType(db *gorm.DB, field *schema.Field) string {
+	return JSONGormDBDataType(db.Dialector.Name())
+}
+
+func (rs ResourceSelectors) GormValue(ctx context.Context, db *gorm.DB) clause.Expr {
+	data, _ := json.Marshal(rs)
+	return gorm.Expr("?", string(data))
```
yashmehrotra, line 155: This is also repeating, lets make this generic as well

types/config.go
```diff
@@ -0,0 +1,61 @@
...
+	if c.Name != "" {
+		s += "/" + c.Name
+	}
+
+	if len(c.Tags) > 0 {
+		s += " " + fmt.Sprintf("%v", c.Tags)
+	}
+
+	return s
+}
```
yashmehrotra, line 42: Lets just use `%+v` or something like that here ?

types/log_selector.go
```diff
@@ -28,15 +28,7 @@ func (t *LogSelectors) Scan(val any) error {
...
-	switch db.Dialector.Name() {
-	case SqliteType:
-		return JSONType
-	case PostgresType:
-		return JSONBType
-	case SQLServerType:
-		return NVarcharType
-	}
-	return ""
+	return JSONGormDBDataType(db.Dialector.Name())
```
yashmehrotra, line 31: 🚀

utils/hash.go
```diff
@@ -0,0 +1,22 @@
...
+		return "", err
+	}
+
+	hash := md5.Sum(data)
+	if err != nil {
+		return "", err
+	}
+
+	return hex.EncodeToString(hash[:]), nil
+}
```
yashmehrotra, line 22: Please this to commons/hash, this snippet is used/repeated in other repos too

models/components.go
```diff
@@ -97,40 +89,100 @@ func (component Component) GetAsEnvironment() map[string]interface{} {
...
 
+func (component Component) GetID() string {
+	if component.ID != uuid.Nil {
+		return component.ID.String()
+	}
+	if component.Text != "" {
+		return component.Text
+	}
+	return component.Name
+}
```
yashmehrotra, line 134: Is this function used anywhere ? Ideally it should always return id and throw error on it being nil

models/components.go
yashmehrotra: Hmm, not sure about it. @moshloop Can you chime in here

types/config.go
adityathebe: interestingly, %+v for map makes no difference.


```
fmt.Printf("%v\n", x)
fmt.Printf("%+v\n", x)
fmt.Printf("%#v\n", x)

map[address:global name:flanksource]
map[address:global name:flanksource]
map[string]string{"address":"global", "name":"flanksource"}
```

I think just `%v` looks fine.

utils/hash.go
adityathebe: cool. will try to move as much utils from other repos as well to commons.

adityathebe: https://github.com/flanksource/commons/pull/72

models/components.go
adityathebe: I see that it's only used in the `Debug` method. https://github.com/flanksource/duty/blob/4fd7266159f67e96ff1cbd179a5ef9a7daf5cbe0/models/components.go#L207

yashmehrotra: Ohh, we can delete this method then

types/config.go
yashmehrotra: Lets just `%+v` or `%v` for the entire String method, since we'll need all the attributes when printing it in logs

PR #140, opened 2023-05-02
fix: topology query

adityathebe:
* teams in topology query
* log selectors in topology view

### DIFF
--- a/views/010_topology.sql
+++ b/views/010_topology.sql
@@ -45,23 +45,10 @@ WITH
       LEFT JOIN teams ON team_components.team_id = teams.id
     GROUP BY
       team_components.component_id
-  ),
-  log_selectors AS (
-    SELECT
-      c.id AS component_id,
-      jsonb_agg(s.selector ->> 'name') AS logs
-    FROM
-      components c,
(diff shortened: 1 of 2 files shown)


views/010_topology.sql
```diff
@@ -48,6 +48,7 @@ WITH
   )
 SELECT
   components.*,
+  components.log_selectors AS logs,
```
moshloop, line 51: I think we just want the log selector name

views/010_topology.sql
```diff
@@ -45,9 +45,23 @@ WITH
       LEFT JOIN teams ON team_components.team_id = teams.id
     GROUP BY
       team_components.component_id
+  ),
+  log_selectors AS (
```
moshloop, line 49: Can we not use jsonb_path_query for this ?

adityathebe: changed it to use `jsonb_path_query`.

But turns out the formed `logs` was incorrect. We currently have `["log-selector", "log-selector-2"]` but we want `[{"name": "log-selector"}, {"name": "log-selector-2"}]`

PR #145, opened 2023-05-10
added test columns in check & check_status tables

adityathebe:


models/checks.go
```diff
@@ -21,31 +21,32 @@ var CheckHealthStatuses = []CheckHealthStatus{
...
+	Labels             types.JSONStringMap  `json:"labels" gorm:"type:jsonstringmap"`
+	Description        string               `json:"description,omitempty"`
+	Status             CheckHealthStatus    `json:"status,omitempty"`
+	Uptime             Uptime               `json:"uptime"  gorm:"-"`
+	Latency            Latency              `json:"latency"  gorm:"-"`
+	Statuses           []CheckStatus        `json:"checkStatuses"  gorm:"-"`
+	Owner              string               `json:"owner,omitempty"`
+	Severity           string               `json:"severity,omitempty"`
+	Icon               string               `json:"icon,omitempty"`
+	TestThreshold      *types.TestThreshold `json:"testThreshold,omitempty" gorm:"column:test_threshold"`
```
moshloop, line 40: Shouldn't this be part of spec ?

moshloop: I think we should split the types into 2, 1 for the spec, and 1 for the results , The results wouldn't have any reference to threshold - it would only have a severity, and likewsie the spec wouldn't have status, uptime, latency, etc..

@yashmehrotra wdyt

yashmehrotra: I'd prefer splitting due to separation of concerns

We keep `summary` in db, but never use that field, it always get calculated at runtime, should look into fixing that sometime as well

PR #146, opened 2023-05-10
fix logs in topology view

adityathebe:


adityathebe:
Runs fine locally

image

PR #152, opened 2023-05-12
feat: components with log selector view

adityathebe:
Resolves: https://github.com/flanksource/duty/issues/150

### DIFF
--- a/views/005_component_views.sql
+++ b/views/005_component_views.sql
@@ -82,6 +82,20 @@ ORDER BY
   key,
   value;
 
+CREATE OR REPLACE VIEW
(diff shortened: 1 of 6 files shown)


views/005_component_views.sql
```diff
@@ -82,6 +82,20 @@ ORDER BY
...
+SELECT
+  id,
+  name,
+  icon,
+type
+FROM
+  components
+WHERE
+  deleted_at IS NULL
+  AND agent_id IS NOT NULL
```
moshloop, line 96:
```suggestion
  AND agent_id IS NULL
```

PR #155, opened 2023-05-14
fix: add agent_id in canaries unique constraint

yashmehrotra:
Fixes: https://github.com/flanksource/incident-commander/issues/342


moshloop:
@yashmehrotra  - can we add tests for this - I don't think postgres handles unique constraints on null's very well

PR #168, opened 2023-05-29
feat: add scraper id to config analysis

adityathebe:
* The default queries using config analysis filters only open analyses

### DIFF
--- a/models/config.go
+++ b/models/config.go
@@ -25,6 +25,13 @@ const (
 	ConfigClassVirtualMachine = "VirtualMachine"
 )
 
+// Config Analysis statuses
+const (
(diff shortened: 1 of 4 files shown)


models/config.go
```diff
@@ -166,6 +173,7 @@ type ConfigAnalysis struct {
 	ExternalID    string        `gorm:"-"`
 	ConfigType    string        `gorm:"-"`
 	ConfigID      uuid.UUID     `gorm:"column:config_id;default:''" json:"config_id"`
+	ScraperID     *uuid.UUID    `gorm:"column:scraper_id;default:null" json:"scraper_id"`
```
moshloop, line 176: Should we not be defaulting the status to open?

PR #174, opened 2023-06-07
feat: added component id to incidents

adityathebe:


adityathebe:
We can get the component id from the evidence

PR #186, opened 2023-06-27
refactor: logic for pushing changes into event queue

yashmehrotra:


views/012_changelog.sql
```diff
@@ -7,7 +7,9 @@ DECLARE
 BEGIN
   rec = NEW;
   IF TG_OP = 'DELETE' THEN
-    rec = OLD;
+    -- TODO: Deletions are not handled
+    -- Handle deletions in reconcile and do not push them in queue
+    return OLD;
```
yashmehrotra, line 12: Right now deletions are not handled by event_queue

We can:
1. Ignore deletions and let reconcile handle them
2. In event queue, if an id is present but can't be fetched, so infer that its been deleted and ask the host to delete it as well

Ideally, I think we should do both, @moshloop @adityathebe what are your thoughts ?

moshloop: I am not sure we should be pushing deletes - I think retention settings should cater to this and you may have different retention settings on the client and server

yashmehrotra: Okay, I'll then remove the comment and do not push deletes

PR #194, opened 2023-06-30
fix: support teams responder in incident_summary view

adityathebe:


views/009_incident_views.sql
```diff
@@ -85,17 +85,22 @@ CREATE OR REPLACE VIEW incident_summary AS
...
-      LEFT JOIN people ON responders.person_id = people.id
-    WHERE
-      people.id IS NOT NULL
-    ORDER BY
-      people.id
+      INNER JOIN people ON responders.person_id = people.id
+    UNION
+    SELECT
+      DISTINCT ON (teams.id, responders.incident_id) teams.id as id,
+      '' as avatar, -- Teams do not have avatar.
```
moshloop, line 98: They do have an icon though ?

adityathebe: Ok if the icon name can suffice for all use cases. I assumed, avatars for users were user upload images (URL) where as the icon is just a name.

PR #207, opened 2023-07-07
feat: added new events to the event queue for notification

adityathebe:
* Added new notification table and removed previous one

### DIFF
--- a/schema/notifications.hcl
+++ b/schema/notifications.hcl
@@ -21,7 +21,7 @@ table "notifications" {
   column "properties" {
     null    = true
     type    = jsonb
-    comment = "Shoutrrr properties shared by all the receivers (person, team & custom services)."
+    comment = "Shoutrrr properties used when sending email to the person receipient."
   }
   column "person_id" {
     null    = true
--- a/schema/notifications.hcl
+++ b/schema/notifications.hcl
@@ -39,7 +39,7 @@ table "notifications" {
     comment = "other 3rd party services for the notification like Slack, Telegram, ..."
   }
   column "created_by" {
-    null = false
+    null = true
     type = uuid
   }
   column "created_at" {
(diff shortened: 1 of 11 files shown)


schema/notifications.hcl
```diff
@@ -0,0 +1,34 @@
...
+  }
+  column "template" {
+    null = false
+    type = text
+  }
+  column "receivers" {
+    null    = false
+    type    = jsonb
+  }
+  column "created_at" {
```
moshloop, line 20: need:

* created_by

foreign keys for: as we will need to show notification for a team/person

* team_id
* person_id 

* filter

schema/notifications.hcl
```diff
@@ -0,0 +1,72 @@
...
+    columns     = [column.created_by]
+    ref_columns = [table.people.column.id]
+    on_update   = NO_ACTION
+    on_delete   = NO_ACTION
+  }
+  foreign_key "notification_person_id_fkey" {
+    columns     = [column.created_by]
+    ref_columns = [table.people.column.id]
+    on_update   = NO_ACTION
+    on_delete   = NO_ACTION
```
moshloop, line 64: I think these should probably cascade - When you delete a person or team, their notifications also get removed

moshloop: I think you would specify team or person - noth both

schema/notifications.hcl
```diff
@@ -0,0 +1,77 @@
...
+  }
+  column "filter" {
+    null = true
+    type = text
+  }
+  column "properties" {
+    null    = true
+    type    = jsonb
+    comment = "Shoutrrr properties used when sending email to the person receipient."
+  }
```
adityathebe, line 25: @moshloop I didn't remove this as we still need customization when sending email (like from name and subject)

schema/notifications.hcl
```diff
@@ -0,0 +1,77 @@
...
+  column "deleted_at" {
+    null = true
+    type = timestamptz
+  }
+  foreign_key "notification_created_by_fkey" {
+    columns     = [column.created_by]
+    ref_columns = [table.people.column.id]
+    on_update   = NO_ACTION
+    on_delete   = CASCADE
```
moshloop, line 63: I don't think this should cascade - Just because a person is deleted, doesn't mean that the notifications they have created for other teams/people should also

PR #216, opened 2023-07-17
feat: disabled checks

adityathebe:


schema/checks.hcl
```diff
@@ -319,3 +319,14 @@ table "check_statuses_1d" {
     columns = [column.created_at]
   }
 }
+
+table "disabled_checks" {
```
moshloop, line 323: @adityathebe  We should be using the properties table for this

PR #221, opened 2023-07-18
feat: added check summary for component func

adityathebe:
Resolves: https://github.com/flanksource/canary-checker/issues/1144

### DIFF
--- a/views/013_check_summary.sql
+++ b/views/013_check_summary.sql
@@ -47,15 +47,6 @@ EXECUTE PROCEDURE update_last_transition_time_for_check ();
 
 -- check summary view
 CREATE OR REPLACE VIEW check_summary AS
-  WITH check_component_relationship_by_check AS (
-  SELECT
-    check_id,
-    json_agg(component_id) AS components
-  FROM
-    check_component_relationships
-  GROUP BY
-    check_id
-  )
   SELECT
     checks.id,
     checks.canary_id,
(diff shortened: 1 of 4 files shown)


views/013_check_summary.sql
```diff
@@ -44,3 +44,63 @@ CREATE
 OR REPLACE TRIGGER checks_last_transition_time BEFORE
 UPDATE ON checks FOR EACH ROW WHEN (OLD.status IS DISTINCT FROM NEW.status)
 EXECUTE PROCEDURE update_last_transition_time_for_check ();
+
+-- check summary view
+DROP VIEW IF EXISTS check_summary;
+
+CREATE VIEW check_summary AS
```
adityathebe, line 51: @moshloop the canary checker's `/api/summary` endpoint can simply call this.
Added it here instead so it's also usable in `check_summary_for_component`

moshloop: @adityathebe  we cannot include components in the default check_summary, You could have some checks linked to hundreds or thousands of components - especially when considering we might want to walk up or down the component hierarchy - i.e. Select a component and then show all health checks for related components.

adityathebe: @moshloop This is the same query that we've been using so far https://github.com/flanksource/duty/blob/main/check_summary.go#L16-L56. 
I just created a view out of it for reusability purpose.

Let me confirm if the `components` column is currently being used by the frontend.

moshloop: Can we also add:

*  check_summary_for_component_down : functions to return all checks below a component i.e. `path like component.path.%` 
*  check_summary_for_component_up : functions to return all checks above a component i.e. `where component_id in split(path, '.')`
*  check_summary_for_component_all: combination of up and down

PR #238, opened 2023-08-04
feat: add access token table

adityathebe:


models/people.go
```diff
@@ -30,3 +32,15 @@ func (p PersonProperties) Value() (driver.Value, error) {
 func (p *PersonProperties) Scan(val any) error {
 	return types.GenericStructScan(&p, val)
 }
+
+type AccessToken struct {
+	ID        uuid.UUID `json:"id" gorm:"default:generate_ulid()"`
```
moshloop, line 37: I think add a name field

schema/system.hcl
```diff
@@ -1,3 +1,41 @@
...
+  column "value" {
+    null = false
+    type = text
+  }
+  column "created_at" {
+    null = false
+    type = timestamptz
+  }
+  column "expires_at" {
+    null = true # We can have never expiring tokens
```
adityathebe, line 21: Should we have never expiring tokens? @moshloop

moshloop: I think we can handle this later

PR #245, opened 2023-08-07
feat: add unique email index in people table

adityathebe:
https://github.com/flanksource/incident-commander/issues/487


moshloop:
How does this handle people without emails e.g. agents ?

PR #249, opened 2023-08-11
feat: introduce dummy data generator that assigns random ID to the models

adityathebe:
Introduces new way to get the same dummy fixtures but with random IDs. This way we can generate as many dummy fixtures as we want for multiple agents.

The old fixtures with hardcoded ids are still there.

### DIFF
--- a/fixtures/dummy/all.go
+++ b/fixtures/dummy/all.go
@@ -3,125 +3,282 @@ package dummy
 import (
 	"fmt"
 
+	"github.com/flanksource/duty/models"
+	"github.com/google/uuid"
 	"gorm.io/gorm"
 )
 
-func PopulateDBWithDummyModels(gormDB *gorm.DB) error {
+type DummyData struct {
+	People                       []models.Person
+	Agents                       []models.Agent
+	Topologies                   []models.Topology
+	Components                   []models.Component
+	ComponentRelationships       []models.ComponentRelationship
+	Configs                      []models.ConfigItem
+	ConfigChanges                []models.ConfigChange
+	ConfigAnalyses               []models.ConfigAnalysis
+	ConfigComponentRelationships []models.ConfigComponentRelationship
+	Teams                        []models.Team
+	Incidents                    []models.Incident
+	Hypotheses                   []models.Hypothesis
+	Evidences                    []models.Evidence
+	Canaries                     []models.Canary
+	Checks                       []models.Check
+	CheckStatuses                []models.CheckStatus
+	Responders                   []models.Responder
+	Comments                     []models.Comment
+	CheckComponentRelationships  []models.CheckComponentRelationship
+}
+
+// GenerateDummyData generates a set of dummy data.
+// If randomize is true, the IDs of the data will be randomly generated.
+func GenerateDummyData(randomize bool) DummyData {
+	// we're appending here so we do not mutate the original slice.
+	d := DummyData{
+		People:                       append([]models.Person(nil), AllDummyPeople...),
+		Agents:                       append([]models.Agent(nil), AllDummyAgents...),
+		Topologies:                   append([]models.Topology(nil), AllDummyTopologies...),
+		Components:                   append([]models.Component(nil), AllDummyComponents...),
+		ComponentRelationships:       append([]models.ComponentRelationship(nil), AllDummyComponentRelationships...),
+		Configs:                      append([]models.ConfigItem(nil), AllDummyConfigs...),
+		ConfigChanges:                append([]models.ConfigChange(nil), AllDummyConfigChanges...),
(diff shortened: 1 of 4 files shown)


fixtures/dummy/all.go
```diff
@@ -130,114 +169,872 @@ func PopulateDBWithDummyModels(gormDB *gorm.DB) error {
...
+		Labels:     types.JSONStringMap{"telemetry": "enabled"},
+		Owner:      "logistics-team",
+		CreatedAt:  DummyCreatedAt,
+		Status:     types.ComponentStatusHealthy,
+	}
+
+	var LogisticsAPI = models.Component{
+		ID:         uuid.New(),
+		Name:       "logistics-api",
+		ExternalId: "dummy/logistics-api",
```
yashmehrotra, line 395: Wouldn't this lead to us having to update fixtures in 2 places ?

yashmehrotra: For all the resources

yashmehrotra:
@adityathebe I still strongly suggest we should just add specific fixtures in a `dummy/agent_aws` and `dummy/agent_gcp` package, they solve the simple purpose to:
1. Compartmentalize resources per different agents.
2. Only be required when testing host <-> agent flow
3. Easy to maintain in the long run

The maintainability overhead does not justify the use case assigning IDs to dummy models

adityathebe:
@yashmehrotra what this PR has is basically what you're suggesting except there's no package per agent but rather a single generator func `GetStaticDummyData` https://github.com/flanksource/duty/pull/249/files#diff-6a84001bcc2a74122c7993b5fc0300078b0ace6693204c8257bf439d935abe1eR290

It achieves all the things that new dummy data package per agent allows with an added flexibility of not having to modify duty if we want to add more agents to the test in incident-commander repo for example.

fixtures/dummy/all.go
adityathebe: We don't have to update the fixture in 2 places. The old dummy data and ones in this func are completely isolated and can grow up/down independently.

yashmehrotra:
@adityathebe But duplicating static data in multiple places will always create a drift

fixtures/dummy/all.go
yashmehrotra: Then why use duplicated values ? This just feels confusing & needlessly complicated

adityathebe:
@yashmehrotra 

> But duplicating static data in multiple places will always create a drift
Can you elaborate on this? For our use case, we only care about the ids and those are not static.

> If we are going down this route, why not just use a function to replace the ids ?
Do you mean to remove all the dummy fixtures and just use the new generator func with an option to randomize the ids or not?

Or 

that we only have a single fixture and then have a func to replace the ids if needed? I've already tried this & abandoned it as the solution turned out to be very messy.

yashmehrotra:
>>    But duplicating static data in multiple places will always create a drift

> Can you elaborate on this? For our use case, we only care about the ids and those are not static.

Basically, if the fixtures with fixed ID and fixtures with static ID are independent of each other, they should use different names and a different persona.
Right now the fixtures are written in a way to reflect a real world scenario (logistics service, cart api, databases, cluster components etc.)

If the fixtures with dynamic IDs can mature independently, we should then use different names for components, configs etc. ie. if they are independent of them, they should also not be derived of them

For fixtures, the rule of thumb is that they should be accurately able to represent the desired state of our database.

For upstream test, the actual problem we want to solve is, test the flow by having N agents and 1 host. We need to test whether the data on the host is the exact same as agents or not

The reason for this change is that we are trying to insert the same data into different agents and then push that to the host and the IDs collide because of being static.

Now instead of modifying IDs, we would want to create explicit static fixtures that represent the same state so that we not only test this flow, but also test all other flows where agent is involved

For eg. when creating the topology tree or generating the check status summary, the database should have unique dummy data for different agents to test how the application behaves with their presence as that reflects how the app will be running in production

Last few weeks, most of the bugs I was fixing were related to this, resources of agents being processed where as they shouldn't. That would have been avoided if we had data populated with agents so we could test the exclusion correctly as well

adityathebe:
The fixture with dynamic ids are still representing a real world scenario since it's just a copy.

> if they are independent of them, they should also not be derived of them

I wouldn't say they are derived. They are copies but still independent. 

> Last few weeks, most of the bugs I was fixing were related to this, resources of agents being processed where as they shouldn't. That would have been avoided if we had data populated with agents so we could test the exclusion correctly as well

But we had agent populated data in the upstream. I think we simply weren't testing that at all.

moshloop:
@adityathebe  / @yashmehrotra -Going to merge this and created a follow-up issue to decide on a test/dummy data strategy

PR #250, opened 2023-08-14
feat: sync team_components table with the deletion of components & teams

adityathebe:
Resolves: https://github.com/flanksource/incident-commander/issues/408


views/007_events.sql
```diff
@@ -195,4 +199,21 @@ $$ LANGUAGE plpgsql;
...
\ No newline at end of file
+EXECUTE PROCEDURE notifications_trigger_function ();
+
+-- Handle component updates
+CREATE
+OR REPLACE FUNCTION handle_component_updates () RETURNS TRIGGER AS $$
+BEGIN
+  IF OLD.deleted_at IS NULL AND NEW.deleted_at IS NOT NULL THEN
+    DELETE FROM team_components WHERE component_id = OLD.id;
+  END IF;
```
yashmehrotra, line 210: Can we also do deletions for component relationships and other related resources as well ?

yashmehrotra: Also, I think these should be in a different file ? Like how we created a new one for canaries/check deletion handling

Doesn't belong in events.sql

yashmehrotra (approved):
:shipit:

views/018_component_triggers.sql
```diff
@@ -0,0 +1,22 @@
+CREATE
+OR REPLACE FUNCTION handle_component_updates () RETURNS TRIGGER AS $$
+BEGIN
+  IF OLD.deleted_at IS NULL AND NEW.deleted_at IS NOT NULL THEN
+    DELETE FROM team_components WHERE component_id = OLD.id;
```
moshloop, line 5: I am not sure we should be doing this - If I am looking at a deleted component, I still want to see what it was related to before it was deleted. 

A relationship should only be deleted when it is no longer true e.g. An EBS volume is detached from an ec2instance - in which case the component/config relationship.deleted_at should reflect the detach time

views/019_teams_triggers.sql
```diff
@@ -0,0 +1,15 @@
+CREATE
+OR REPLACE FUNCTION handle_team_updates () RETURNS TRIGGER AS $$
+BEGIN
+  IF OLD.deleted_at IS NULL AND NEW.deleted_at IS NOT NULL THEN
+    DELETE FROM team_components WHERE team_id = OLD.id;
```
moshloop, line 5: This is ok, when deleting a team they no longer "own" any components.

views/018_component_triggers.sql
moshloop: @adityathebe  / @yashmehrotra  ^

moshloop (changes requested):

moshloop (approved):

PR #252, opened 2023-08-15
feat: playbooks

adityathebe:


models/playbooks.go
```diff
@@ -0,0 +1,27 @@
...
+import (
+	"time"
+
+	"github.com/flanksource/duty/types"
+	"github.com/google/uuid"
+)
+
+type Playbook struct {
+	ID          uuid.UUID  `gorm:"default:generate_ulid()"`
+	Description string     `json:"description"`
```
moshloop, line 12:
```suggestion
	Name string     `json:"name"`
```

models/playbooks.go
```diff
@@ -0,0 +1,27 @@
...
+	CreatedAt   time.Time  `json:"created_at,omitempty" time_format:"postgres_timestamp"`
+	UpdatedAt   time.Time  `json:"updated_at,omitempty" time_format:"postgres_timestamp"`
+	DeletedAt   *time.Time `json:"deleted_at,omitempty" time_format:"postgres_timestamp"`
+}
+
+type PlaybookRun struct {
+	ID          uuid.UUID  `gorm:"default:generate_ulid()"`
+	PlaybookID  uuid.UUID  `json:"playbook_id"`
+	CreatedAt   time.Time  `json:"created_at,omitempty" time_format:"postgres_timestamp"`
+	StartedAt   *time.Time `json:"started_at,omitempty" time_format:"postgres_timestamp"`
```
moshloop, line 24: This could be scheduled for the future so I think StartDate and EndDate is better.

Also need references to:

* Component / config executed on
* Agent 

And columns for
* Parameters 
* Results 
* Status (Scheduled, Running, Cancelled, Completed, Failure)

models/playbooks.go
```diff
@@ -0,0 +1,44 @@
...
+	ID          uuid.UUID         `gorm:"default:generate_ulid()"`
+	PlaybookID  uuid.UUID         `json:"playbook_id"`
+	CreatedAt   time.Time         `json:"created_at,omitempty" time_format:"postgres_timestamp" gorm:"default:NOW()"`
+	StartDate   time.Time         `json:"start_date,omitempty" time_format:"postgres_timestamp" gorm:"default:NOW()"`
+	EndDate     *time.Time        `json:"end_date,omitempty" time_format:"postgres_timestamp"`
+	Duration    time.Duration     `json:"duration" gorm:"default:null"`
+	Result      types.JSON        `json:"result,omitempty"`
+	CreatedBy   *uuid.UUID        `json:"created_by,omitempty"`
+	ComponentID *uuid.UUID        `json:"component_id,omitempty"`
+	ConfigID    *uuid.UUID        `json:"config_id,omitempty"`
```
yashmehrotra, line 40: Can a run have multiple of these ids ?

adityathebe: A run can have only one of either config or the component id.

models/playbooks.go
```diff
@@ -0,0 +1,56 @@
...
+}
+
+type PlaybookRun struct {
+	ID          uuid.UUID           `gorm:"default:generate_ulid()"`
+	PlaybookID  uuid.UUID           `json:"playbook_id"`
+	Status      PlaybookRunStatus   `json:"status,omitempty"`
+	CreatedAt   time.Time           `json:"created_at,omitempty" time_format:"postgres_timestamp" gorm:"default:NOW()"`
+	StartTime   time.Time           `json:"start_time,omitempty" time_format:"postgres_timestamp" gorm:"default:NOW()"`
+	EndTime     *time.Time          `json:"end_time,omitempty" time_format:"postgres_timestamp"`
+	Duration    time.Duration       `json:"duration" gorm:"default:null"`
```
moshloop, line 38: I don't think we need to keep duration as we can calculate it

PR #254, opened 2023-08-21
feat: add external id in people table

yashmehrotra:


schema/teams.hcl
```diff
@@ -61,6 +64,10 @@ table "people" {
     unique  = true
     columns = [column.email]
   }
+  index "people_external_id_idx" {
+    unique  = true
```
moshloop, line 68: I don't think this can be unique if we have null values ?

PR #260, opened 2023-08-28
feat: add sync events

adityathebe:


views/007_events.sql
```diff
@@ -37,6 +37,7 @@ CREATE OR REPLACE FUNCTION insert_responder_in_event_queue() RETURNS TRIGGER AS
 BEGIN
     IF TG_OP = 'INSERT' THEN
         INSERT INTO event_queue(name, properties) VALUES ('incident.responder.added', jsonb_build_object('id', NEW.id));
+        INSERT INTO event_queue(name, properties) VALUES ('incident.responder.added.sync', jsonb_build_object('id', NEW.id));
```
moshloop, line 40: Why do we need this? Why would any of the incident.responder.added events be handled async?

adityathebe: It's notifying 3rd party responder clients: Jira & MSPlanner.

views/007_events.sql
```diff
@@ -123,6 +125,26 @@ AFTER UPDATE ON checks
...
 
+-- Insert component status updates in event_queue
+CREATE OR REPLACE FUNCTION insert_component_status_updates_in_event_queue() RETURNS TRIGGER AS $$
+DECLARE event_name TEXT;
+BEGIN
+    IF OLD.status = NEW.status THEN
+      RETURN NULL;
+    END IF;
+
+    event_name := 'component.' || NEW.status || '.sync';
```
adityathebe, line 136: @moshloop Unlike checks, components have a variety of statuses so the event name is created this way.

```
const (
	ComponentStatusHealthy   ComponentStatus = "healthy"
	ComponentStatusUnhealthy ComponentStatus = "unhealthy"
	ComponentStatusWarning   ComponentStatus = "warning"
	ComponentStatusError     ComponentStatus = "error"
	ComponentStatusInfo      ComponentStatus = "info"
)
```

PR #271, opened 2023-09-08
feat: add cached getters and modify `AsMap()` to accept fields to remove

adityathebe:


getters.go
```diff
@@ -0,0 +1,73 @@
...
+			return cache.resource.(*T), nil
+		} else {
+			logger.Warnf("Unexpected cached value type: %T", value)
+		}
+	}
+
+	var resource T
+	if err := ctx.DB().Where("id = ?", id).First(&resource).Error; err != nil {
+		if errors.Is(err, gorm.ErrRecordNotFound) {
+			getterCache.SetDefault(id, cachedItem{notFound: true})
```
moshloop, line 64: I don't think we should be caching not founds

getters.go
```diff
@@ -0,0 +1,73 @@
...
+	"time"
+
+	"github.com/flanksource/commons/logger"
+	"github.com/flanksource/duty/models"
+	"github.com/patrickmn/go-cache"
+	"gorm.io/gorm"
+)
+
+// getterCache caches the results for all the getters in this file.
+var getterCache = cache.New(time.Second*30, time.Minute*5)
```
moshloop, line 14:
```suggestion
var getterCache = cache.New(time.Second*90, time.Minute*5)
```

getters.go
```diff
@@ -0,0 +1,73 @@
...
+
+type cachedItem struct {
+	resource any
+	notFound bool
+}
+
+func FindCachedAgent(ctx dbContext, agentID string) (*models.Agent, error) {
```
moshloop, line 21: Can we also add for Person, Component and Config?

Why is the first parameter a private struct ?

getters.go
```diff
@@ -0,0 +1,73 @@
...
+func findCachedEntity[T any](ctx dbContext, id string) (*T, error) {
+	if value, ok := getterCache.Get(id); ok {
+		if cache, ok := value.(cachedItem); ok {
+			if cache.notFound {
+				return nil, nil
+			}
+
+			return cache.resource.(*T), nil
+		} else {
+			logger.Warnf("Unexpected cached value type: %T", value)
```
moshloop, line 57: This should be an error

getters.go
adityathebe: added Person, Component and Config.

The first parameters is an interface.

moshloop: Yes, but why is it private ?

adityathebe: I don't see why that should be exported.

moshloop: We should also handle the special case of all 000-s agent and return nil

moshloop: > I don't see why that should be exported. 

The types of the parameters for exported function should all be exported, otherwise you may not be able to call the function, or the type info gets lost in docs etc..

adityathebe: As interfaces are implicit, callers outside of the package will still be able to call the function. And the callers can still see what needs to be defined to satisfy the interface.

From a usability perspective, I still do not see the difference. Moreover, exporting it also enables other importers to internally use it _(eg: config-db using it internally without any calls to duty)_. Any future change to `dbContext` might cause the consumers to break.

I would lean more towards minimizing the export footprint of the package and only exporting the must-haves.

The only downside I see is that it might not be well documented with `go doc`.

PR #277, opened 2023-09-13
feat: add more getters

adityathebe:


getters.go
```diff
@@ -14,66 +14,84 @@ import (
...
 	if err != nil {
 		return nil, err
 	}
 
 	return canary, nil
 }
 
-func FindCachedPerson(ctx DBContext, agentID string) (*models.Person, error) {
-	person, err := findCachedEntity[models.Person](ctx, agentID)
+func FindCachedPerson(ctx DBContext, id string) (*models.Person, error) {
```
moshloop, line 48: Why can't we just have FindPerson and FindTeam and then it checks whether it is a uuid, string or email ?

adityathebe: Do we want to have the option to optionally not use cache? I think cached getter is mainly for notification templating.

PR #281, opened 2023-09-15
feat: new notification send history table

adityathebe:


models/notifications.go
```diff
@@ -32,3 +32,34 @@ func (n *Notification) HasRecipients() bool {
 func (n Notification) AsMap(removeFields ...string) map[string]any {
 	return asMap(n, removeFields...)
 }
+
+type NotificationSendHistory struct {
+	ID             uuid.UUID `json:"id,omitempty" gorm:"default:generate_ulid()"`
+	NotificationID string    `json:"notification_id"`
+	Body           string    `json:"body,omitempty"`
+	Error          *string   `json:"error,omitempty"`
+	DurationMs     int64     `json:"duration_ms,omitempty" gorm:"column:duration_millis"`
```
moshloop, line 41:
```suggestion
	DurationMillis     int64     `json:"duration_ms,omitempty" gorm:"column:duration_millis"`
```

schema/notifications.hcl
```diff
@@ -81,4 +81,63 @@ table "notifications" {
...
+    default = sql("now()")
+  }
+  primary_key {
+    columns = [column.id]
+  }
+  foreign_key "notification_id_fkey" {
+    columns     = [column.notification_id]
+    ref_columns = [table.notifications.column.id]
+    on_update   = NO_ACTION
+    on_delete   = NO_ACTION
```
moshloop, line 135: I think these should cascade

moshloop:
@adityathebe  can we also create Cleanup function to remove notifications older than 30d ?

.github/workflows/test.yaml
```diff
@@ -51,7 +51,7 @@ jobs:
 
     services:
       postgres:
-        image: postgres
+        image: postgres:15
```
moshloop, line 54: Why the bump ?

schema/notifications.hcl
```diff
@@ -81,4 +81,63 @@ table "notifications" {
...
+    type    = timestamptz
+    default = sql("now()")
+  }
+  primary_key {
+    columns = [column.id]
+  }
+  foreign_key "notification_id_fkey" {
+    columns     = [column.notification_id]
+    ref_columns = [table.notifications.column.id]
+    on_update   = NO_ACTION
```
moshloop, line 134: Shouldn't these also cascade ?

.github/workflows/test.yaml
adityathebe: The latest is 16 so I was actually trying to downgrade it. For some reason, migration failed with v16.
What's our preferred version?

moshloop: It looks like all major clouds now support 15, but if we bump it from 14 to 15 here, we also need to bump it everywhere else

adityathebe: > but if we bump it from 14 to 15 here

I think this Github action has been running on **15** for quite some time now since it was the default.
But if we're running on 14 everywhere else, I'll tag this to 14 here as well.

PR #284, opened 2023-09-21
feat: support reconciliation for a specified duration in the past

adityathebe:
- [x] Removes trigger on check statuses table to insert push_queue events to the event_queue.
- [x] Use RFC3339Nano instead of RFC3339
- [x] Removed pg utils. Moved to `flanksource/postq`


upstream/reconcile.go
```diff
@@ -43,14 +47,55 @@ func NewUpstreamReconciler(upstreamConf UpstreamConfig, pageSize int) *upstreamR
...
 		next = "," // in the format <check_id>,<time>
 	}
 
+	return t.sync(ctx, table, next)
+}
+
+// Sync compares all the resource of the given table against
+// the upstream server and pushes any missing resources to the upstream.
+func (t *upstreamReconciler) SyncAfter(ctx duty.DBContext, table string, after time.Duration) error {
+	logger.WithValues("since", time.Now().Add(-after).Format(time.RFC3339)).Debugf("Reconciling table %q with upstream", table)
```
moshloop, line 65: We need to use the uuid for comparison, created_at could be along time/before or after the db insertion time. 
On startup request the max uuid, and use that for the after.

adityathebe: I am not sure I follow. We're still using UUID for comparison throughout the reconciliation loop _(in `sync` function)_. This is just one call where we fetch the row to start the reconciliation from.

moshloop: The insertion time and the created date are not the same - I could bring in something with a nil created_at or a created_at from years ago and then those entries will not be synchronised - There are also risks of timezone and clock skew also causing issues. 

On the client side of the push, on startup we should do a `select max(id) from table` and then do a sync after that id.

adityathebe: `created_at` that we're using is the db insertion time.
The record's created at date is the `time` column.

So I think we're covering all new check statuses inserted in the last 48 hours.

adityathebe: NOTE: created_at is not reliable (example: configs)

PR #287, opened 2023-09-22
feat: add error column to notifications table

adityathebe:
* setup new pg notifications when tables are updated

### DIFF
--- a/views/007_events.sql
+++ b/views/007_events.sql
@@ -218,27 +218,6 @@ WHERE
 GROUP BY
   name;
 
--- Insert team updates in event_queue
-CREATE
-OR REPLACE FUNCTION insert_team_in_event_queue () RETURNS TRIGGER AS $$
-BEGIN
-  IF TG_OP = 'DELETE' THEN
-    INSERT INTO event_queue(name, properties) VALUES ('team.delete', jsonb_build_object('team_id', OLD.id))
-    ON CONFLICT (name, properties) DO UPDATE SET created_at = NOW(), last_attempt = NULL, attempts = 0;
-  ELSE
-    INSERT INTO event_queue(name, properties) VALUES ('team.update', jsonb_build_object('team_id', NEW.id))
-    ON CONFLICT (name, properties) DO UPDATE SET created_at = NOW(), last_attempt = NULL, attempts = 0;
-  END IF;
-  
-  RETURN NULL;
-END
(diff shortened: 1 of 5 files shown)


views/007_events.sql
```diff
@@ -218,48 +218,6 @@ WHERE
 GROUP BY
   name;
 
--- Insert team updates in event_queue
-CREATE
-OR REPLACE FUNCTION insert_team_in_event_queue () RETURNS TRIGGER AS $$
```
adityathebe, line 223: Removed this from being pushed to the event queue because the event consumer in IC were simply clearing the `teams` & `responder` cache. https://github.com/flanksource/incident-commander/blob/1265484533b60bfb1196c017e9769dfc9151a617/events/teams.go#L34-L70

Instead now we have PG Notify on the teams table.

views/007_events.sql
```diff
@@ -218,48 +218,6 @@ WHERE
...
-
-    RETURN NULL;
-END
-$$ LANGUAGE plpgsql;
-
-CREATE OR REPLACE TRIGGER notification_update_enqueue
-AFTER INSERT OR UPDATE OR DELETE ON notifications
-FOR EACH ROW
-EXECUTE PROCEDURE notifications_trigger_function ();
-
```
adityathebe, line 262: Removed this event for the same reason. The consumer was simply clearing notification cache.
It's now moved to PG Notify listener.

views/020_teams_triggers.sql
```diff
@@ -1,14 +1,26 @@
+-- Notify on any updates on the teams table
 CREATE OR REPLACE FUNCTION handle_team_updates()
 RETURNS TRIGGER AS $$
 BEGIN
+  IF TG_OP = 'DELETE' THEN
+    PERFORM pg_notify('table_activity', TG_TABLE_NAME || ' ' || OLD.id);
```
adityathebe, line 6: Created a new NOTIFY channel `table_activity` which should be used to publish any table updated notify instead of having multiple notify channels like `teams_updated` or `notifications_updated`.

This way we can use the notify router and have one listener on the NOTIFY channel and then distribute to multiple consumers based on the table name.

adityathebe: The payload should be in the format `<route> <an optional payload>`. The route is used by the pg notify router and the remaining payload is sent to the consumers.

Issue #285, opened 2023-09-20
Topology query - don't return details children 


moshloop:
Beyond the direct children of components we should only return summary e.g. on https://mission-control.azure.lab.flanksource.com/topology/0189f817-9bb9-a8bc-4a1e-c19830134576?sortBy=name&sortOrder=asc
nodes and pods should have all details but individual nodes and pods should only have name, namespace, id, is_leaf, status, status_reason, summary, topology_type, agent and type

PR #288, opened 2023-09-25
feat: minimal topology details for rest of the children

adityathebe:
Resolves:  https://github.com/flanksource/duty/issues/285

### DIFF
--- a/topology.go
+++ b/topology.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"encoding/json"
 	"fmt"
+	"slices"
 	"strings"
 
 	"github.com/flanksource/commons/collections"
@@ -26,7 +27,8 @@ type TopologyOptions struct {
 	Types  []string
 	Status []string
 
-	// when set to true, only the children (except the immediate children) are returned
+	// when set to true, only the children (except the immediate children) are returned.
+	// when set to false, the direct children & the parent itself is fetched.
 	restOfTheChildrenOnly bool
 }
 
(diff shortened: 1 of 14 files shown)


moshloop:
@adityathebe  can we add some tests for this ?

topology.go
```diff
@@ -25,17 +26,43 @@ type TopologyOptions struct {
...
 
+// selectClause returns the columns that should be selected from the topology view.
+func (opt TopologyOptions) selectClause() string {
+	if !opt.restOfTheChildrenOnly {
+		return "*"
+	}
+
+	// parents, incidents, analysis, checks columns need to fetched to create the topology tree even though they may not be essential to the UI.
+	return "name, namespace, id, is_leaf, status, status_reason, summary, topology_type, labels, team_names, agent_id, parent_id, type, parents, incidents, analysis, checks"
+}
```
adityathebe, line 47: @moshloop had to add back these columns as they're essential during tree creation.
- parents
- incidents
- analysis
- checks

moshloop: I am not sure how much better this will perform - I also think we should use more sql functions / views and have this in a single db query rather than multiple trips to the db.

Generally I think there are phases:

1. Find all top level components (no parents, or direct descendants) and apply label, type, team, agent filters at this level only. 
2. For each component summarize health, incidents, analysis of its descendants 
    
```
// find all the descendants of a component, including those via relationships
find_descendants(id): returns id array
````

```
descendants =  find_descendants(id)
select components.* from components where filters
with each top level component
select name, id, parent, type, icon, component_summary.* from components where id in @descandents 
inner join component_summary
```

where component_summary is a view, that maybe becomes a materialized view in the future (which is refreshed on a debonce(check.transitioned, analysis.updated, incident.transitioned) etc..

3. Union and then post process for top-level summary in go
- strip off parent(s), agent_id in final topology returned 

wdyt @yashmehrotra  ?

File: topology.go [hunk 2]
```diff
@@ -321,3 +398,31 @@ func GetComponent(ctx context.Context, db *gorm.DB, id string) (*models.Componen
...
+			m[v] = struct{}{}
+			a = append(a, v)
+		}
+	}
+
+	sort.Slice(a, func(i, j int) bool {
+		return a[i] < a[j]
+	})
+	return a
+}
```
yashmehrotra, line 428: use collections.Set in commons for this

topology.go
yashmehrotra: Agree with point 1 of applying filters just at top level

for point 2, are we selecting all the descendants or just summarizing them ? cause summary right now happens in go not sql, so we might want to change that

topology.go
```diff
@@ -25,17 +26,43 @@ type TopologyOptions struct {
 	// TODO: Filter status and types in DB Query
 	Types  []string
 	Status []string
+
+	// when set to true, only the children (except the immediate children) are returned.
+	// when set to false, the direct children & the parent itself is fetched.
+	restOfTheChildrenOnly bool
```
yashmehrotra, line 32: what do we mean by immediate, direct and rest of the children ?

The terminology is very confusing

PR #290, opened 2023-09-27
feat: add agents summary view with count of agents

mainawycliffe:


views/022_agents.sql
```diff
@@ -0,0 +1,18 @@
+CREATE OR REPLACE VIEW
+    agents_summary AS
+SELECT
```
moshloop, line 3: I don't think this query will be very efficient, its better to use a sub-query e.g.

select agents.*, configs.count as config_count from agents inner join (select agent_id, count(id) as count from config_items group by agent_id) as configs on configs.agent_id = agents.id

mainawycliffe: Something like this?

PR #297, opened 2023-10-10
feat: add shared context

moshloop:
Depends on https://github.com/flanksource/commons/pull/81


context/context.go
```diff
@@ -0,0 +1,139 @@
...
+	"github.com/flanksource/duty"
+	"github.com/flanksource/duty/models"
+	"github.com/flanksource/duty/types"
+
+	"k8s.io/client-go/kubernetes"
+)
+
+type Context struct {
+	*commons.Context
+}
```
adityathebe, line 20: What happens with `duty.DBContext`?

adityathebe: Where would this be used? The public facing duty APIs cannot have a type constraint.

moshloop: DBContext falls away,  and in theory none of the projects would need to create their own context - especially if we move the echo context here as well

> public facing duty APIs cannot have a type constraint.

Not sure we are using a type constraint, or why it would be an issue?

adityathebe: > DBContext falls away, and in theory none of the projects would need to create their own context - especially if we move the echo context here as well

I think echo context isn't even used right now. We just use the request context from it so I think that can go away.

> Not sure we are using a type constraint, or why it would be an issue?

In mission control I wouldn't be able to pass in the context ( `api.Context`) if duty's public facing api are accepting `context.Context` struct assuming we're going to be using this new `context.Context` struct inside duty.

moshloop: In mission control you don't need a api.Context, you can use duty.Context directly, if we do not a custom context then we create a new context that reuses the same context in the same way duty.Context and commons.Context share the same context

adityathebe: > if we do not a custom context then we create a new context that reuses the same context in the same way duty.Context and commons.Context share the same context

Assuming we'll want to keep the custom context structs in
- **canary-checker** which has the canary spec & kommons, and
- **config-db** which has the scraper spec

, simply passing ctx when calling db functions wouldn't work. 

So, I think we need to export the context interface

PR #300, opened 2023-10-12
feat: add check statuses syncing func

adityathebe:
This will be used by incident-commander & canary-checker

### DIFF
--- a/upstream/jobs.go
+++ b/upstream/jobs.go
@@ -11,45 +11,35 @@ import (
 
 // SyncCheckStatuses pushes check statuses, that haven't already been pushed, to upstream.
 func SyncCheckStatuses(ctx duty.DBContext, config UpstreamConfig, batchSize int) error {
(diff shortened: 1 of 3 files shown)


upstream/jobs.go
```diff
@@ -0,0 +1,55 @@
...
+
+	"github.com/flanksource/commons/logger"
+	"github.com/flanksource/duty"
+	"github.com/flanksource/duty/models"
+	"github.com/google/uuid"
+)
+
+// SyncCheckStatuses pushes check statuses, that haven't already been pushed, to upstream.
+func SyncCheckStatuses(ctx duty.DBContext, config UpstreamConfig, batchSize int) error {
+	var checkStatuses []models.CheckStatus
```
moshloop, line 14: I think the the batching should occur here - There could be potentially tens of thousands of statuses to replicate.

PR #302, opened 2023-10-16
feat: add job runtime and job struct

yashmehrotra:
Depends on https://github.com/flanksource/commons/pull/84

### DIFF
--- a/context/context.go
+++ b/context/context.go
@@ -1,6 +1,7 @@
 package context
 
 import (
+	gocontext "context"
 	commons "github.com/flanksource/commons/context"
 	"github.com/jackc/pgx/v5/pgxpool"
 	"go.opentelemetry.io/otel/attribute"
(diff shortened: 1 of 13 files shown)


job/job.go
```diff
@@ -0,0 +1,77 @@
...
+	Name            string
+	Schedule        string
+	AllowConcurrent bool
+	Timeout         time.Duration
+	Fn              func(ctx JobRuntime) error
+	RunNow          bool
+}
+
+type JobRuntime struct {
+	context.Context
```
yashmehrotra, line 23: I think this should be a field `Context context.Context` since we are passing `.Context` of this in all the funcs that use context.Context like db funcs

context/context.go
```diff
@@ -71,6 +84,11 @@ func (k Context) PgxPool() *pgxpool.Pool {
 	return k.Value("pgxpool").(*pgxpool.Pool)
```
adityathebe, line 84: I guess `PgxPool() *pgxpool.Pool` can be removed in favor of `Pool`

context/context.go
```diff
@@ -137,3 +155,11 @@ func (k *Context) HydratedConnectionByURL(namespace, connectionString string) (*
 func (k *Context) HydrateConnection(connection *models.Connection, namespace string) (*models.Connection, error) {
 	return duty.HydrateConnection(k, k.Kubernetes(), k.DB(), connection, namespace)
 }
+
+func WrapContext(gormDB *gorm.DB, pool *pgxpool.Pool, k8s kubernetes.Interface, tracer trace.Tracer) func(gocontext.Context) Context {
```
moshloop, line 159:
```suggestion
func (k *Context) Wrap( ctx gocontext.Context) func(t) Context {
 return NewContext(ctx, commons.WithTracer(k.Tracer()).WithDB(k.DB(), k.Pool()).WithKubernetes(k.Kubernetes())
}
```

yashmehrotra: @moshloop  Don't we want a closure ? or do we want to keep a default context and clone that with new context ?

PR #303, opened 2023-10-18
feat: add AsGoGetterURL() to connection

adityathebe:


models/connections.go
```diff
@@ -48,3 +93,79 @@ func (c Connection) String() string {
...
+		}
+
+		parsedURL.RawQuery = q.Encode()
+		output = parsedURL.String()
+	}
+
+	return output, nil
+}
+
+func (c Connection) AsEnv() []string {
```
moshloop, line 154: This needs to return a list of files to create as well e.g.  for gcp

models/connections.go
```diff
@@ -9,6 +10,50 @@ import (
 	"github.com/google/uuid"
 )
 
+// List of all connection types
+const (
```
moshloop, line 14: I thunk these are lowercase

adityathebe: These are all saved as uppercase (& some with spaces). It's coming from https://github.com/flanksource/flanksource-ui/blob/main/src/components/Connections/connectionTypes.tsx


```
 SELECT DISTINCT type FROM connections;
+--------------+
| type         |
|--------------|
| Gotify       |
| Zulip Chat   |
| Slack        |
| Google Chat  |
| postgres     |
| Google Cloud |
| AWS          |
| Azure        |
| Discord      |
| Matrix       |
| Email        |
| HTTP         |
| Ntfy         |
| Azure Devops |
| Postgres     |
| Telegram     |
+--------------+
```

models/connections.go
adityathebe: Files need templating & the template variables are not consistent with exec connection. Eg: for aws: https://github.com/flanksource/canary-checker/blob/78fa5e0f54a4cbf75ace4402a0459f5d271cee7e/checks/exec.go#L183-L187

We get the token like `{{.SessionToken.ValueStatic}}` but for connection we'd get the token from properties. So, the files config, I think, can stay with the checks?

models/connections.go
adityathebe: Although, I do think that these needs to be changed so we only have lowercase and no-spaces.

moshloop: Logged https://github.com/flanksource/flanksource-ui/issues/1448 for that

models/connections.go
moshloop: The templating should be done here, as a consumer we get back the hydrated environment variables and files.

models/connections.go
```diff
@@ -48,3 +97,159 @@ func (c Connection) String() string {
...
+}
+
+type EnvPrep struct {
+	Conn Connection
+
+	// Env is the connection credentials in environment variables
+	Env []string
+
+	// File contains the content of the configuration file based on the connection
+	File bytes.Buffer
```
moshloop, line 199: This needs to be map[string]File - consumers are not aware of the type connection and where the file needs to go

models/connections.go
```diff
@@ -48,3 +97,159 @@ func (c Connection) String() string {
...
+
+	case ConnectionTypeAzure:
+		args := []string{"login", "--service-principal", "--username", c.Conn.Username, "--password", c.Conn.Password}
+		if v, ok := c.Conn.Properties["tenant"]; ok {
+			args = append(args, "--tenant")
+			args = append(args, v)
+		}
+
+		// login with service principal
+		runCmd := exec.CommandContext(ctx, "az", args...)
```
moshloop, line 237: We need to return this as part of envprep - Apply should not be conenction type aware as we may use it to apply elsewhere e.g. a kubernetes pod

models/connections.go
```diff
@@ -1,14 +1,63 @@
...
 	"regexp"
 	"time"
 
 	"github.com/flanksource/duty/types"
 	"github.com/google/uuid"
 )
 
+// List of all connection types
+const (
+	ConnectionTypeAWS            = "AWS"
```
moshloop, line 19: I think lower_case these, and then toLower().replaceAll(" ", "_") the input for comparison

models/connections.go
```diff
@@ -48,3 +97,159 @@ func (c Connection) String() string {
...
+	case ConnectionTypeHTTP:
+		if c.Username != "" || c.Password != "" {
+			parsedURL.User = url.UserPassword(c.Username, c.Password)
+		}
+
+		output = parsedURL.String()
+
+	case ConnectionTypeGit:
+		q := parsedURL.Query()
+		q.Set("sshkey", c.Certificate)
```
moshloop, line 120: I think this needs to be base64 encoded

models/connections.go
```diff
@@ -48,3 +102,153 @@ func (c Connection) String() string {
...
+		Files: make(map[string]bytes.Buffer),
+	}
+
+	switch strings.ReplaceAll(strings.ToLower(c.Type), " ", "_") {
+	case ConnectionTypeAWS:
+		envPrep.Env = append(envPrep.Env, fmt.Sprintf("AWS_ACCESS_KEY_ID=%s", c.Username))
+		envPrep.Env = append(envPrep.Env, fmt.Sprintf("AWS_SECRET_ACCESS_KEY=%s", c.Password))
+
+		// credentialFilePath :="$HOME/.aws/credentials"
+		credentialFilePath := filepath.Join(".creds", "aws", fmt.Sprintf("cred-%d", rand.Intn(100000000)))
```
adityathebe, line 178: @moshloop using `fs/FS` wasn't possible because it seems to be read-only. Can't create new files or write to one.

- https://github.com/golang/go/issues/45757

models/connections.go
```diff
@@ -48,3 +102,166 @@ func (c Connection) String() string {
...
+
+		envPrep.CmdEnvs = append(envPrep.CmdEnvs, fmt.Sprintf("GOOGLE_APPLICATION_CREDENTIALS=%s", credentialFilePath))
+	}
+
+	return envPrep
+}
+
+type EnvPrep struct {
+	// Env is the connection credentials in environment variables
+	Env []string
```
moshloop, line 232: What is the difference between envs and command envs ?

adityathebe: Env would provide you the credentials as standalone env vars. You could use either env var or the config file.

PR #311, opened 2023-10-23
chore: add missing deleted_at and cleanup fields to agents table

mainawycliffe:


schema/agents.hcl
```diff
@@ -68,6 +68,17 @@ table "agents" {
     default = sql("now()")
   }
 
+  column cleanup {
```
yashmehrotra, line 71: add quotes

PR #321, opened 2023-11-03
fix: add deduplication in topology

yashmehrotra:


query/topology.go
```diff
@@ -90,9 +92,9 @@ func (opt TopologyOptions) componentRelationWhereClause() string {
 		}
 	} else {
 		if !opt.nonDirectChildrenOnly {
-			s += " AND component_relationships.component_id = NULL"
+			s += " AND component_relationships.component_id IS NULL"
 		} else {
-			s += " AND component_relationships.component_id IS NOT NULL"
+			s += " AND component_relationships.relationship_id IS NULL"
 		}
```
yashmehrotra, line 98: @adityathebe Let me know what these should be

tests/fixtures/expectations/topology_child_tree.json
```diff
@@ -44,38 +44,6 @@
...
-            "healthy": 1
-          },
-          "type": "KubernetesPod"
-        },
-        {
-          "agent_id": "00000000-0000-0000-0000-000000000000",
-          "icon": "icon-kubernetes-pod",
-          "id": "018681ff-b6c1-a14d-2fd4-8c7dac94cddd",
-          "is_leaf": false,
-          "name": "logistics-ui-676b85b87c-tjjcp",
```
yashmehrotra, line 69: @moshloop After the direct children change, there might be bugs/regressions which were copied over in some fixtures as well

For eg. here components are being repeated

PR #330, opened 2023-11-07
feat: check summary filter by check labels

adityathebe:
related: https://github.com/flanksource/canary-checker/issues/1384


query/check_summary.go
```diff
@@ -23,10 +23,11 @@ type CheckSummarySortBy string
...
 type CheckSummaryOptions struct {
-	Timeout    time.Duration
-	CheckID    *uuid.UUID
-	SortBy     CheckSummarySortBy
-	DeleteFrom *time.Time
+	Timeout     time.Duration
+	CheckID     *uuid.UUID
+	SortBy      CheckSummarySortBy
+	DeleteFrom  *time.Time
+	CheckLabels map[string]string
```
moshloop, line 30: These would just be labels that filter over both canary and check labels

adityathebe:
@moshloop I'm not sure if anything needs to be done on this view - https://github.com/flanksource/duty/blob/0276f3188f40837643ea5b430151961b8ff507fb/views/004_check_views.sql#L17-L30. 

Don't see that being used anywhere:

views/013_check_summary.sql
```diff
@@ -87,6 +87,7 @@ CREATE OR REPLACE VIEW check_summary AS
     canaries.namespace,
     canaries.name as canary_name,
     canaries.labels,
+    checks.labels as check_labels,
```
moshloop, line 90: Can we merge these instead ? check labels toake precedence

Issue #333, opened 2023-11-08
Artifacts


moshloop:
Build a generic artifact archiving mechanism into duty 

e.g. 

Artifact {
check_id, check_time,
connection_id,
playbook_run_id, 
filename, size, checksum,created_at,deleted_at,updated_at
}

* Expose a function for artifact_by_check and by playbook run
* Extract the Filesystem interface from canary_checker folder checks and add FilesystemReadWrite interface that extends it
* Add a Connection.AsFilesystem() 
* Move the AWS/GCP/SMB/SFTP connection types to duty, leveraging the Connection.Merge for population

PR #339, opened 2023-11-20
feat: Artifacts

adityathebe:
- https://github.com/flanksource/duty/issues/333


models/artifacts.go
```diff
@@ -0,0 +1,24 @@
...
+
+	"github.com/google/uuid"
+)
+
+// Artifact represents the artifacts table
+type Artifact struct {
+	ID            uuid.UUID  `json:"id"`
+	CheckID       uuid.UUID  `json:"check_id,omitempty"`
+	CheckTime     time.Time  `json:"check_time,omitempty" time_format:"postgres_timestamp"`
+	PlaybookRunID uuid.UUID  `json:"playbook_run_id,omitempty"`
```
moshloop, line 14: Maybe we should use join tables instead e.g. check_artifacts, playbook_run_artifacts

models/artifacts.go
```diff
@@ -0,0 +1,24 @@
...
+	ConnectionID  uuid.UUID  `json:"connection_id,omitempty"`
+	FileID        string     `json:"file_id"`
```
moshloop, line 16: What is the difference between this and id ?

adityathebe: _FileID_ would be the id of the file on the store and id is just the database row id.
eg: For s3, it would be the object id

models/artifacts.go
adityathebe: as discussed we can handle this later

query/artifacts.go
```diff
@@ -0,0 +1,21 @@
...
+
+import (
+	"time"
+
+	"github.com/flanksource/duty/context"
+	"github.com/flanksource/duty/models"
+	"github.com/google/uuid"
+)
+
+func ArtifactsByCheck(ctx context.Context, checkID uuid.UUID, checkTime time.Time) ([]models.Artifact, error) {
```
moshloop, line 11: Can we do these as views, so that they are ready for use by the UI ?  We would not return the connectionId and server side path

adityathebe: This is just a WHERE clause on the artifacts table. Do we need a view for this?

Issue #350, opened 2023-11-29
Add support for  cloudsql connection strings 


moshloop:
This would be for both the main connection to gorm/pool db and for canaries/etc...

https://cloud.google.com/sql/docs/postgres/samples/cloud-sql-postgres-databasesql-auto-iam-authn 

`db=database user=service-account-name@project-id.iam  cloudsql-instance-connection-name=project:region:instance private_ip=true`

PR #353, opened 2023-12-01
feat: support cloudsql connection string

adityathebe:
resolves: https://github.com/flanksource/duty/issues/350


db.go
```diff
@@ -44,6 +51,12 @@ func NewGorm(connection string, config *gorm.Config) (*gorm.DB, error) {
 }
 
 func NewDB(connection string) (*sql.DB, error) {
+	if strings.Contains(connection, "cloudsql-instance-connection-name") {
```
moshloop, line 54: could we move this to a drivers.Parse(url) (url, DialFn) - so that it is reusabe ?

moshloop: @adityathebe ^

Issue #356, opened 2023-12-07
Integrations view


moshloop:
Combine scrapers, topology and loogging backends into a Integrations view with a Integrations type.

PR #362, opened 2023-12-11
feat: integrations view

adityathebe:
resolves: https://github.com/flanksource/duty/issues/356

### DIFF
--- a/schema/components.hcl
+++ b/schema/components.hcl
@@ -25,7 +25,7 @@ table "topologies" {
   column "source" {
     null = false
     type = enum.source
-    default = "UI"
+    default = "Topology"
   }
   column "spec" {
     null = true
--- a/schema/components.hcl
+++ b/schema/components.hcl
@@ -22,6 +22,11 @@ table "topologies" {
     null = true
     type = jsonb
   }
+  column "source" {
+    null = false
+    type = enum.source
+    default = "UI"
+  }
   column "spec" {
     null = true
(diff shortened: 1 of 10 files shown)


views/024_integrations.sql
```diff
@@ -0,0 +1,4 @@
+CREATE OR REPLACE VIEW integrations_todo AS
```
adityathebe, line 1: We seem to already have a "integrations" table
https://github.com/flanksource/duty/blob/987e8c512cbc1235aa145f802c955608a6fb2a39/schema/system.hcl#L93-L139

schema/components.hcl
```diff
@@ -22,6 +22,11 @@ table "topologies" {
     null = true
     type = jsonb
   }
+  column "source" {
+    null = false
+    type = enum.source
+    default = "UI"
```
moshloop, line 28:
```suggestion
    default = "Topology"
```

views/024_integrations.sql
```diff
@@ -0,0 +1,4 @@
+CREATE OR REPLACE VIEW integrations_list AS
+SELECT id, name, description, 'scrapers' AS integration_type, source, agent_id, created_at, updated_at, deleted_at, created_by FROM config_scrapers_with_status UNION
```
moshloop, line 2: We need all the job_history columns as well

Can we also add a test that calls the view ?

PR #364, opened 2023-12-18
fix: reconcile all upstream resources from beginning

yashmehrotra:


upstream/reconcile.go
```diff
@@ -56,12 +56,17 @@ func (t *upstreamReconciler) Sync(ctx context.Context, table string) error {
...
+	//if err := ctx.DB().Table(table).Select("id").Where("agent_id = ?", uuid.Nil).Where("NOW() - updated_at > ?", after).Order("id").Limit(1).Scan(&next).Error; err != nil {
+	//return err
+	//}
+	//if err := ctx.DB().Table(table).Select("id").Where("agent_id = ?", uuid.Nil).Where("NOW() - updated_at > ?", after).Order("id").Limit(1).Scan(&next).Error; err != nil {
+	//return err
+	//}
+
+	// We start with a nil UUID and calculate hash in batches
+	next := uuid.Nil.String()
 	return t.sync(ctx, table, next)
```
adityathebe, line 70: I think we just get rid of `SyncAfter` since now it's identical to `Sync` method above it?

upstream/reconcile.go
```diff
@@ -76,7 +81,6 @@ func (t *upstreamReconciler) sync(ctx context.Context, table, next string) error
 		if err != nil {
 			return fmt.Errorf("failed to fetch hash of primary keys from local db: %w", err)
 		}
-		next = localStatus.Next
```
adityathebe, line 79: Don't we need this in here though instead of inside `if upstreamStatus.Hash == localStatus.Hash`?

The pagination would be stuck in a loop otherwise.

upstream/reconcile.go
yashmehrotra: I want to make sure this works before refactoring

Will clean it up once we decide to go this way 👍🏽

upstream/reconcile.go
yashmehrotra: We can add a counter maybe ? currently we do not check the hash after adding missing resources

adityathebe (approved):

Issue #346, opened 2023-11-26
Add onceoff migration scripts


moshloop:
e.g. with a migrations table that tracks scripts that have been run and skips them

PR #367, opened 2023-12-19
feat: log migration of scripts

adityathebe:
resolves: https://github.com/flanksource/duty/issues/346

### DIFF
--- a/migrate/migrate.go
+++ b/migrate/migrate.go
@@ -2,6 +2,7 @@ package migrate
 
 import (
 	"context"
+	"crypto/sha1"
 	"database/sql"
 	"errors"
 	"fmt"
@@ -33,8 +34,11 @@ func RunMigrations(pool *sql.DB, connection string, opts MigrateOptions) error {
 	if err := row.Scan(&name); err != nil {
 		return fmt.Errorf("failed to get current database: %w", err)
 	}
-
 	logger.Infof("Migrating database %s", name)
+
+	if err := createMigrationLogTable(pool); err != nil {
+		return fmt.Errorf("failed to create migration log table: %w", err)
+	}
 
 	logger.Tracef("Getting functions")
 	funcs, err := functions.GetFunctions()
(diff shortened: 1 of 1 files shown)


migrate/migrate.go
```diff
@@ -148,12 +152,44 @@ func runScripts(pool *sql.DB, scripts map[string]string, ignoreFiles []string) e
 		filenames = append(filenames, name)
 	}
 	sort.Strings(filenames)
+
 	for _, file := range filenames {
+		content, ok := scripts[file]
+		if !ok {
+			continue
+		}
```
yashmehrotra, line 160: Should we not raise error here ?

migrate/migrate.go
```diff
@@ -148,12 +152,44 @@ func runScripts(pool *sql.DB, scripts map[string]string, ignoreFiles []string) e
...
+
+		var currentHash string
+		if err := pool.QueryRow("SELECT hash FROM migration_logs WHERE path = $1", file).Scan(&currentHash); err != nil && !errors.Is(err, sql.ErrNoRows) {
+			return err
+		}
+
+		hash := sha1.Sum([]byte(content))
```
yashmehrotra, line 167:
```suggestion
		hash := sha1.Sum([]byte(content))[:]
```

to avoid doing `hash[:]` below

migrate/migrate.go
```diff
@@ -148,12 +152,44 @@ func runScripts(pool *sql.DB, scripts map[string]string, ignoreFiles []string) e
...
+func createMigrationLogTable(pool *sql.DB) error {
+	query := `CREATE TABLE IF NOT EXISTS migration_logs (
+		path VARCHAR(255) NOT NULL,
+		hash bytea NOT NULL,
+		updated_at TIMESTAMPTZ DEFAULT NOW() NOT NULL,
+		PRIMARY KEY (path)
+	)`
+	_, err := pool.Exec(query)
+	return err
+}
```
yashmehrotra, line 195: Why aren't we using the tables/*.hcl way for this ?

adityathebe: Atlas schemas are run after running the `functions` scripts. And this table needs to be there even before we run any scripts.

migrate/migrate.go
adityathebe: `filenames` are generated from scripts so I don't think we will ever have `ok==false`

migrate/migrate.go
yashmehrotra: can't we do `functions` -> `schema` -> then scripts

adityathebe: That's how it is right now. We keep track of migration of `functions` as well

PR #369, opened 2023-12-19
fix: add bucket to aws connection

adityathebe:


connection/aws.go
```diff
@@ -11,6 +11,7 @@ type AWSConnection struct {
 	// ConnectionName of the connection. It'll be used to populate the endpoint, accessKey and secretKey.
 	ConnectionName string       `yaml:"connection,omitempty" json:"connection,omitempty"`
 	AccessKey      types.EnvVar `yaml:"accessKey" json:"accessKey,omitempty"`
+	Bucket         string       `yaml:"bucket,omitempty" json:"bucket,omitempty"`
```
moshloop, line 14: Shouldn't we use an S3Connection that extends AWSConnection ?

adityathebe: I saw that the existing aws connection struct had some S3 params so I just attached it there. I've added the two new connection structs now.

connection/s3.go
```diff
@@ -0,0 +1,60 @@
...
+
+	"github.com/flanksource/commons/collections"
+)
+
+// +kubebuilder:object:generate=true
+type S3Connection struct {
+	AWSConnection `yaml:",inline"`
+	Bucket        string `yaml:"bucket,omitempty" json:"bucket,omitempty"`
+	// glob path to restrict matches to a subset
+	ObjectPath string `yaml:"objectPath,omitempty" json:"objectPath,omitempty"`
```
adityathebe, line 15: I'm not entirely sure if this should be part of the connection spec or if it's something that's specific to the s3 canary check.

PR #376, opened 2023-12-21
chore: add artifacts_with_check_status view

mainawycliffe:


views/025_artifacts.sql
```diff
@@ -0,0 +1,14 @@
+CREATE OR REPLACE VIEW
+  artifacts_with_check_status AS
+SELECT
+  artifacts.*,
+  check_statuses.status AS check_status,
+  check_statuses.message AS check_message,
+  check_statuses.error AS error,
+  check_statuses.invalid AS invalid,
+  check_statuses.details AS details
+FROM
```
moshloop, line 10: This needs to be the other way around, select from check statuses, joining on an json_aggregation of artifacts so that we get 1 row per check status, with a json array of artifact name, size and id 

The grouping and join is also on check_id + time, which also need to be returned with duration

The details column should not be returned from this view (to access it (maybe via a details link in the row, it should be queried for a specific check_id and time as it can be large - sometimes in MB's

moshloop: Yes, but need to join on time as well

PR #384, opened 2023-12-24
chore: do skip entire check summary test

yashmehrotra:
@moshloop I think we should just ignore those fields which don't match and try to fix them at a later time instead of skipping the whole test


moshloop:
> @moshloop I think we should just ignore those fields which don't match and try to fix them at a later time instead of skipping the whole test

Then we should have a 2nd test rather, the TODO/FIXME is not really visible

PR #407, opened 2024-01-02
feat: add is_pushed columns to changes and analyses

adityathebe:


upstream/jobs.go
```diff
@@ -21,25 +21,95 @@ func SyncCheckStatuses(ctx context.Context, config UpstreamConfig, batchSize int
...
+		logger.Tracef("pushing %d config_changes to upstream", len(configChanges))
+		if err := client.Push(ctx, &PushData{AgentName: config.AgentName, ConfigChanges: configChanges}); err != nil {
+			return 0, fmt.Errorf("failed to push config_changes to upstream: %w", err)
+		}
+
+		for i := range configChanges {
+			configChanges[i].IsPushed = true
+		}
+
+		if err := ctx.DB().Save(&configChanges).Error; err != nil {
```
moshloop, line 75: This should probablt be an update only columns ?

upstream/jobs.go
```diff
@@ -21,25 +21,97 @@ func SyncCheckStatuses(ctx context.Context, config UpstreamConfig, batchSize int
...
+			Limit(batchSize).
+			Find(&configChanges).Error; err != nil {
+			return 0, fmt.Errorf("failed to fetch config_changes: %w", err)
+		}
+
+		if len(configChanges) == 0 {
+			return count, nil
+		}
+
+		logger.Tracef("pushing %d config_changes to upstream", len(configChanges))
```
moshloop, line 66:
```suggestion
		ctx.Tracef("pushing %d config_changes to upstream", len(configChanges))
```

schema/config.hcl
```diff
@@ -58,6 +58,12 @@ table "config_analysis" {
     null = true
     type = timestamptz
   }
+  column "is_pushed" {
```
moshloop, line 61: I think we might need to add a trigger on config_analysis to reset is_pushed when the row is updated (e.g. status or last_observed)

Issue #398, opened 2023-04-19
Resource selector does not use name


adityathebe:
When linking components using selector, the name field is ignored.
`GetComponentsWithSelectors` should also use name.

PR #409, opened 2024-01-02
fix: component resource selector

adityathebe:
resolves: https://github.com/flanksource/duty/issues/398


adityathebe:
Shouldn't the `fieldSelector` be renamed to `propertySelector` @moshloop

moshloop:
> Shouldn't the `fieldSelector` be renamed to `propertySelector` @moshloop

Field selector is what is used in Kubernetes and what to try and keep the semantics the same

PR #416, opened 2024-01-03
fix: add interval to canaries_with_status

adityathebe:
resolves: https://github.com/flanksource/duty/issues/412


moshloop:
Need schedule as well

PR #419, opened 2024-01-04
fix: merge component check status json

yashmehrotra:


moshloop:
@yashmehrotra  can you check the test failures ?

PR #436, opened 2024-01-15
fix: add missing category field to playbook spec

mainawycliffe:


schema/playbooks.hcl
```diff
@@ -29,6 +29,10 @@ table "playbooks" {
     null = false
     type = enum.source
   }
+  column "category" {
+    null = false
```
moshloop, line 33:
```suggestion
    null = true
```

PR #467, opened 2024-01-25
feat: upstream push for artifacts

yashmehrotra:
Fixes https://github.com/flanksource/mission-control/issues/754

### DIFF
--- a/schema/artifacts.hcl
+++ b/schema/artifacts.hcl
@@ -5,6 +5,11 @@ table "artifacts" {
     type    = uuid
     default = sql("generate_ulid()")
   }
+  column "agent_id" {
+    null    = true
+    default = var.uuid_nil
+    type    = uuid
+  }
   column "check_id" {
     null = true
     type = uuid
@@ -13,7 +18,7 @@ table "artifacts" {
     null = true
     type = timestamptz
   }
-  column "playbook_run_id" {
+  column "playbook_run_action_id" {
     null = true
     type = uuid
   }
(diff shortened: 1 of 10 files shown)


schema/artifacts.hcl
```diff
@@ -69,8 +79,8 @@ table "artifacts" {
     on_update   = NO_ACTION
     on_delete   = NO_ACTION
   }
-  foreign_key "artifacts_playbook_run_fkey" {
-    columns     = [column.playbook_run_id]
+  foreign_key "artifacts_playbook_run_action_fkey" {
+    columns     = [column.playbook_run_action_id]
     ref_columns = [table.playbook_runs.column.id]
```
adityathebe, line 84: The table would be `playbook_run_actions`

schema/artifacts.hcl
```diff
@@ -5,6 +5,11 @@ table "artifacts" {
     type    = uuid
     default = sql("generate_ulid()")
   }
+  column "agent_id" {
```
adityathebe, line 8: do we need agent id in here? Since artifacts would just be linked to a check/playbook action that already contains the agent id?

schema/artifacts.hcl
yashmehrotra: 👍🏽

schema/artifacts.hcl
yashmehrotra: 👍🏽

PR #470, opened 2024-01-26
feat: add a limited view for playbook spec

mainawycliffe:


views/018_playbooks.sql
```diff
@@ -76,4 +76,16 @@ SELECT
 FROM
   interim
   INNER JOIN agents ON interim.agent_name :: TEXT = agents.name
-GROUP BY agent_name, agents.person_id, agent_id
\ No newline at end of file
+GROUP BY agent_name, agents.person_id, agent_id;
+
+DROP VIEW IF EXISTS playbook_items;
```
moshloop, line 81: I think use playbook_names, which is the convention we use elsewhere

PR #478, opened 2024-01-30
feat: config traversal

yashmehrotra:


query/config_traversal.go
```diff
@@ -0,0 +1,50 @@
...
+		}
+
+		typeIDs, err := ConfigIDsByTypeFromCache(ctx, relType)
+		if err != nil {
+			return configItem, fmt.Errorf("no type %s exists for any config: %w", relType, err)
+		}
+
+		configID, ok := lo.Find(relatedIDs, func(relID string) bool {
+			return lo.Contains(typeIDs, relID)
+		})
```
moshloop, line 34: I think this is going to be really slow, can we not cache it directly e.g. 

`{id}/Kubernetes::Node []string`

query/config_traversal.go
```diff
@@ -0,0 +1,50 @@
...
+import (
+	"fmt"
+	"strings"
+
+	"github.com/flanksource/duty/context"
+	"github.com/flanksource/duty/models"
+	"github.com/samber/lo"
+)
+
+func TraverseConfig(ctx context.Context, id, relationType string) (models.ConfigItem, error) {
```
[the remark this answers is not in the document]
yashmehrotra: Thought we can keep an error to highlight if part of the traversal is correct and other part wrong, ie. if you do `Deployment/HelmRelease/Kustomize` and `Deplyoment/HelmRelease` is correct but  HelmRelease has no Kustomize then the error message should reflect that

And we can handle what to return in templating in the wrapper func

What do you think ?

query/config_traversal.go
yashmehrotra: Done 👍🏽

query/config_traversal.go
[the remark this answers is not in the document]
yashmehrotra: Okay

query/config_cache.go
```diff
@@ -156,3 +159,84 @@ func FindConfigForComponent(ctx context.Context, componentID, configType string)
...
+	return "configID:" + id
+}
+
+// <config_id> -> []related_ids
+var configRelationCache = cache.New[[]string](gocache_store.NewGoCache(gocache.New(6*time.Hour, 6*time.Hour)))
+
+func configRelationCacheKey(id string) string {
+	return "configRelatedIDs:" + id
+}
+
```
yashmehrotra, line 183: 10 mins for cache time, job runs every 5 mins

moshloop: @yashmehrotra is there a commit missing ?

yashmehrotra: Sorry, forgot to push the commit

Issue #488, opened 2024-02-01
Getters  / Caching improvements


moshloop:
- [x] Use `FindCachedAgent` https://github.com/flanksource/duty/blob/9d4cfd9d442ee092a7ef9d50b312ef17da94b0d7/getters.go#L378
- [x] Remove `FindComponentsByName`
- [x] Use ctx as first arg, not DB
- [x] Update `firstResourceSelectorQuery` to include label and field selectors, these conditions are anded e.g. find components of type X with labels a [ ]bc
- [x] Remove `FindConfigsByField` in favor of ResourceSelector
- [x] Rename `FindConfigsByIDs` etc to to `GetConfigsByID`
- [x] Rename `FindCachedComponent` etc to `GetComponent`
- [x] Add component query cache - selector.Hash -> []uuid.Id
- [x] Switch FindConfig to query/config_cache
- [x] Add ConfigQuery.ToResourceSelector() and standardize the cache on that
- [x] Move the component caching to query/

PR #493, opened 2024-02-01
Getters / Caching improvements

adityathebe:
resolves: https://github.com/flanksource/duty/issues/488


models/config.go
```diff
@@ -96,14 +96,6 @@ func (c ConfigItem) Key() string {
 	return c.ID.String()
 }
 
-func (ConfigItem) LabelsColumn() string {
-	return "tags"
-}
-
-func (ConfigItem) TableName() string {
```
moshloop, line 103: I think this is unintentional

query/config.go
```diff
@@ -2,14 +2,65 @@ package query
...
+
+func FindConfigs(ctx context.Context, config types.ConfigQuery) ([]models.ConfigItem, error) {
+	return FindConfigsByResourceSelector(ctx, config.ToResourceSelector())
+}
+
+func FindConfigIDs(ctx context.Context, config types.ConfigQuery) ([]uuid.UUID, error) {
+	return FindConfigIDsByResourceSelector(ctx, config.ToResourceSelector())
+}
+
+func FindConfigsByResourceSelector(ctx context.Context, resourceSelectors ...types.ResourceSelector) ([]models.ConfigItem, error) {
```
adityathebe, line 40: The naming is a bit awkward because we already have `FindConfigIDs` that operates on `types.ConfigQuery`

moshloop:
@mainawycliffe don;t we need a `make generate` ?

PR #515, opened 2024-02-12
feat: add component name/type/parent/topologyid cache

yashmehrotra:


query/components_cache.go
```diff
@@ -24,25 +25,41 @@ const cacheJobBatchSize = 1000
 // <id> -> models.Component
 var componentCache = cache.New[models.Component](gocache_store.NewGoCache(gocache.New(10*time.Minute, 10*time.Minute)))
 
+// <name/type/parentID/topologyID> -> models.Component
+var componentNameTypeParentTopologyCache = cache.New[models.Component](gocache_store.NewGoCache(gocache.New(30*time.Minute, 30*time.Minute)))
```
yashmehrotra, line 29: @moshloop Is a 30 minute cache too long ? Or should I reduce it

query/components_cache.go
```diff
@@ -24,25 +25,41 @@ const cacheJobBatchSize = 1000
...
+	pID := uuid.Nil.String()
+	if parentID != nil {
+		pID = parentID.String()
+	}
+	return name + typ + pID + topologyID
+}
+
 func SyncComponentCache(ctx context.Context) error {
-	var next string
+	next := uuid.Nil.String()
```
yashmehrotra, line 44: Without this, the query was always failing with `invalid input syntax for type uuid: ""`

query/components_cache.go
```diff
@@ -24,25 +25,41 @@ const cacheJobBatchSize = 1000
...
 	for {
 		var components []models.Component
-		if err := ctx.DB().Where("deleted_at IS NULL").Where("id > ?", next).Limit(cacheJobBatchSize).Find(&components).Error; err != nil {
+		if err := ctx.DB().Where("agent_id = ?", uuid.Nil).Where("id > ?", next).Limit(cacheJobBatchSize).Find(&components).Error; err != nil {
```
adityathebe, line 47: > Where("agent_id = ?", uuid.Nil)

We need to cache components of all agents as well since resource selector uses this cache.

query/components_cache.go
```diff
@@ -24,25 +25,41 @@ const cacheJobBatchSize = 1000
...
+	}
+	return name + typ + pID + topologyID
+}
+
 func SyncComponentCache(ctx context.Context) error {
-	var next string
+	next := uuid.Nil.String()
 	for {
 		var components []models.Component
-		if err := ctx.DB().Where("deleted_at IS NULL").Where("id > ?", next).Limit(cacheJobBatchSize).Find(&components).Error; err != nil {
```
adityathebe, line 35: why do we need to cache deleted components? Is it for the graph thing you're working on?

If we cache them, then we should probably be excluding deleted items from `func GetComponentsByIDs`

yashmehrotra: This is for persisting a component, if a component is deleted but recreated, that case needs to be handled

query/components_cache.go
yashmehrotra: Cool, i'll add agent_id as part of the cacheKey

PR #527, opened 2024-02-14
chore: param include deleted for resource selector

yashmehrotra:


types/resource_selector.go
```diff
@@ -24,6 +25,8 @@ type ResourceSelector struct {
 	//  'max-age=X' (cache for X duration)
 	Cache string `yaml:"cache,omitempty" json:"cache,omitempty"`
 
+	IncludeDeleted bool `yaml:"includeDeleted" json:"includeDeleted"`
```
moshloop, line 28: This should probably be yaml ignored as it doesn't make sense for a user to specify it

yashmehrotra: Done

types/resource_selector.go
```diff
@@ -24,6 +25,8 @@ type ResourceSelector struct {
 	//  'max-age=X' (cache for X duration)
 	Cache string `yaml:"cache,omitempty" json:"cache,omitempty"`
 
+	IncludeDeleted bool `yaml:"-" json:"-"`
```
adityathebe, line 28: should probably be included in `IsEmpty()`

PR #562, opened 2024-03-01
feat: add is_pushed columns to tables for reconciliation

adityathebe:


moshloop:
@adityathebe don't we need is_pushed on the relationships as well ?

adityathebe:
> @adityathebe don't we need is_pushed on the relationships as well ?

I put relationships to the side for now as we talked about evaluating them on the upstream and not having to need to sync them.

upstream/jobs.go
```diff
@@ -7,136 +7,293 @@ import (
...
-// SyncArtifacts pushes artifacts that haven't already been pushed to upstream.
-func SyncArtifacts(ctx context.Context, config UpstreamConfig, batchSize int) (int, error) {
+// ReconcileTable pushes all unpushed items in a table to upstream.
+func reconcileTable[T dbTable](ctx context.Context, config UpstreamConfig, fetcher *gorm.DB, batchSize int) (int, error) {
 	client := NewUpstreamClient(config)
-	count := 0
+	var anon T
+	table := anon.TableName()
+
+	if ctx.Properties()["upstream.reconcile.pre-check"] != "false" {
```
moshloop, line 192:
```suggestion
	if !ctx.Properties().Off("upstream.reconcile.pre-check") {
```

upstream/jobs.go
```diff
@@ -7,136 +7,293 @@ import (
...
+// ReconcileTable pushes all unpushed items in a table to upstream.
+func reconcileTable[T dbTable](ctx context.Context, config UpstreamConfig, fetcher *gorm.DB, batchSize int) (int, error) {
 	client := NewUpstreamClient(config)
-	count := 0
+	var anon T
+	table := anon.TableName()
+
+	if !ctx.Properties().Off("upstream.reconcile.pre-check") {
+		var unpushed float64
+		precheck := fmt.Sprintf(`SELECT reltuples FROM pg_class WHERE relname = '%s_is_pushed_idx'`, table)
```
moshloop, line 194: This query should be part of PushAll run the query to return the tables with rows to push, and then only push those tables.

```
select c.relname from pg_index i
join pg_class c on c.oid = i.indrelid
join pg_class ic on i.indexrelid = ic.oid
WHERE  i.indexrelid::regclass::text like '%_is_pushed_idx' and  ic.reltuples > 0
```

upstream/jobs.go
```diff
@@ -5,138 +5,149 @@ import (
...
+}
 
-		ctx.Tracef("pushing %d config_analyses to upstream", len(analyses))
-		if err := client.Push(ctx, &PushData{ConfigAnalysis: analyses}); err != nil {
-			return 0, fmt.Errorf("failed to push config_analysis to upstream: %w", err)
+func ReconcileSome(ctx context.Context, config UpstreamConfig, batchSize int, runOnly ...string) (int, error) {
+	var count int
+	for _, table := range reconciledTables {
+		if len(runOnly) > 0 && !lo.Contains(runOnly, table.TableName()) {
+			ctx.Tracef("skipping reconciliation of table %s", table.TableName())
```
moshloop, line 59: This shouldn't be logged

PR #580, opened 2024-03-07
feat: related changes search 

adityathebe:
This PR adds 
- `query.FindCatalogChanges` for improved catalog changes searching
- support for **both** in `related_changes_recursive` in addition to the existing `upstream` & `downstream` recursion type.


query/config.go
```diff
@@ -148,3 +153,83 @@ func FindConfigForComponent(ctx context.Context, componentID, configType string)
...
+
+	fromParsed time.Time
+}
+
+func (t *CatalogChangesSearchRequest) Validate() error {
+	if t.CatalogID == uuid.Nil {
+		return fmt.Errorf("catalog id is required")
+	}
+
+	if t.Recursive == "" || (t.Recursive != "upstream" && t.Recursive != "downstream") {
```
moshloop, line 173: recursive is optional, and should also support all or both

PR #583, opened 2024-03-07
feat: add a job_last_failed column for tables with summary views

mainawycliffe:
This is related to this -> https://github.com/flanksource/flanksource-ui/issues/1728

### DIFF
--- a/views/015_job_history.sql
+++ b/views/015_job_history.sql
@@ -38,11 +38,20 @@ SELECT
   job_history_latest_status.status job_status,
   job_history_latest_status.time_start job_time_start,
   job_history_latest_status.time_end job_time_end,
-  job_history_latest_status.created_at job_created_at
+  job_history_latest_status.created_at job_created_at,
+  lr.time_start job_last_failed
 FROM
   topologies
   LEFT JOIN job_history_latest_status ON topologies.id::TEXT = job_history_latest_status.resource_id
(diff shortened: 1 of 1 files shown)


views/015_job_history.sql
```diff
@@ -38,11 +38,20 @@ SELECT
...
 FROM
   topologies
   LEFT JOIN job_history_latest_status ON topologies.id::TEXT = job_history_latest_status.resource_id
   AND job_history_latest_status.resource_type = 'topology'
+  LEFT JOIN (
+    SELECT 
+      resource_id,
+      MAX(time_start) as time_start
+    FROM job_history as js_last_failed
+    WHERE js_last_failed.status = 'failed'
```
adityathebe, line 52:
```suggestion
    WHERE js_last_failed.status = 'FAILED'
```

adityathebe: https://github.com/flanksource/duty/blob/f3e11b7c6937dc513dec1f64db9f44ecdd0e4f15/models/job_history.go#L22

views/015_job_history.sql
```diff
@@ -38,11 +38,20 @@ SELECT
...
+    GROUP BY js_last_failed.resource_id
```
adityathebe, line 53: I think also add this filter in the LEFT JOIN sub query as well.

```
AND js_last_failed.resource_type = 'topology'
```

PR #610, opened 2024-03-14
fix: related_configs_recursive

moshloop:


views/006_config_views.sql
```diff
@@ -610,24 +518,50 @@ BEGIN
...
+    ORDER BY depth ASC;
 END;
 $$ LANGUAGE plpgsql;
 
--- related config changes recursively
-DROP FUNCTION IF EXISTS related_changes_recursive(UUID, TEXT, BOOLEAN);
+-- related configs
+DROP FUNCTION IF EXISTS related_configs;
 
-CREATE FUNCTION related_changes_recursive (
```
adityathebe, line 621: @moshloop Why are we removing this? It's being used by the catalog/changes endpoint
https://github.com/flanksource/duty/blob/fb9b1e136d24775cff357caa92a6d7a7370fe551/query/config_changes.go#L142

Issue #585, opened 2024-03-08
Resource Selectors - Support full kubernetes style selectors 


moshloop:

moshloop:
Using the gorm clause API

+ Add ResourceSelector.Match using 

https://github.com/kubernetes/apimachinery/blob/master/pkg/fields/selector.go and 
https://github.com/kubernetes/apimachinery/blob/master/pkg/labels/labels.go

adityathebe:
related: https://github.com/flanksource/duty/issues/490

PR #635, opened 2024-03-22
feat: use field/label selector parser from apimachinery

adityathebe:
resolves: https://github.com/flanksource/duty/issues/585

### DIFF
--- a/models/checks.go
+++ b/models/checks.go
@@ -9,6 +9,8 @@ import (
 	"github.com/samber/lo"
 	"gorm.io/gorm"
 	"gorm.io/gorm/clause"
+	"k8s.io/apimachinery/pkg/fields"
+	"k8s.io/apimachinery/pkg/labels"
 )
 
 type CheckHealthStatus string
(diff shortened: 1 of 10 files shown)


yashmehrotra:
@adityathebe Is this change backward compatible with current implementation ?

adityathebe:
> @adityathebe Is this change backward compatible with current implementation ?

everything else should work fine apart from the `a=` syntax for `exists` query. That's not supported by the apimachinery's label parser.

yashmehrotra:
> everything else should work fine apart from the `a=` syntax for `exists` query. That's not supported by the apimachinery's label parser.

We use `labelSelector: kustomize.fluxcd.io/name` to fetch Configs which are managed by flux

adityathebe:
@yashmehrotra oh that's the syntax. That should be easy to support. Let me make that change.

adityathebe:
need to make some changes on this PR after [this](https://github.com/flanksource/mission-control/pull/928) is merged.

PR #644, opened 2024-03-27
feat: use directions in config traversal

yashmehrotra:
Fixes: https://github.com/flanksource/duty/issues/640

### DIFF
--- a/query/config_traversal.go
+++ b/query/config_traversal.go
@@ -1,7 +1,6 @@
 package query
 
 import (
-	"github.com/flanksource/commons/logger"
 	"github.com/flanksource/duty/context"
 	"github.com/flanksource/duty/models"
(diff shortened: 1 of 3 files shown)


query/config_traversal.go
```diff
@@ -11,44 +9,50 @@ import (
 	"github.com/google/cel-go/common/types/ref"
 )
 
-func TraverseConfig(ctx context.Context, id, relationType string) *models.ConfigItem {
-	var configItem models.ConfigItem
-
-	relationTypeList := strings.Split(relationType, "/")
+func TraverseConfig(ctx context.Context, id, relationType, direction string) []models.ConfigItem {
```
moshloop, line 12: I think we should still support the traversal path, there might be cases with multiple incoming items have Kustomization parents,

Issue #646, opened 2024-03-28
Config traversal  - Support path


moshloop:
I think we should still support the traversal path, there might be cases with multiple incoming items have Kustomization parents,

_Originally posted by @moshloop in https://github.com/flanksource/duty/pull/644#discussion_r1541560369_

PR #648, opened 2024-03-28
chore: add support for path in config traversal

yashmehrotra:
Fixes: https://github.com/flanksource/duty/issues/646

### DIFF
--- a/query/config_traversal.go
+++ b/query/config_traversal.go
@@ -1,6 +1,8 @@
 package query
 
 import (
+	"strings"
+
 	"github.com/flanksource/duty/context"
 	"github.com/flanksource/duty/models"
 	"github.com/flanksource/gomplate/v3/conv"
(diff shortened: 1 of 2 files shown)


yashmehrotra:
@moshloop The logic can be improved I guess but let me know if this is the direction we want to go for

query/config_traversal.go
```diff
@@ -34,6 +31,27 @@ func TraverseConfig(ctx context.Context, id, relationType, direction string) []m
 	return configItems
 }
 
+// Fetch config IDs which match the type and direction upto max depth (5)
+func getRelatedTypeConfigID(ctx context.Context, ids []string, relatedType, direction string) []string {
+	var allIDs []string
+	for _, id := range ids {
+		q := ctx.DB().Table("related_configs_recursive(?, ?)", id, direction).Select("id", "depth", "type").Where("type = ?", relatedType)
```
yashmehrotra, line 38: @moshloop Using related_configs_recursive function as its max depth is more and allows us to do skip level relation matching

PR #651, opened 2024-04-01
feat: config tags. remove namespace & account columns. rename  tags->labels

adityathebe:


functions/drop.sql
adityathebe, line 1: droping `namespace` column was not possible as a couple of config views depended on it.

Needed a way to drop those views before we ran the atlas schema. So I put it in here.

schema/config.hcl
```diff
@@ -242,7 +234,7 @@ table "config_items" {
     null = true
     type = text
   }
-  column "tags" {
```
adityathebe, line 245: didn't find any way to rename columns using hcl schemas.

query/resource_selector.go
```diff
@@ -56,7 +56,7 @@ func SearchResources(ctx context.Context, req SearchResourcesRequest) (*SearchRe
 				output.Configs = append(output.Configs, SelectedResource{
 					ID:        items[i].GetID(),
 					Agent:     items[i].AgentID.String(),
-					Labels:    lo.FromPtr(items[i].Tags),
+					Labels:    lo.FromPtr(items[i].Labels),
```
moshloop, line 59: We should be returning tags here, not labels

views/006_config_views.sql
```diff
@@ -487,7 +520,7 @@ CREATE FUNCTION related_configs_recursive (
     direction TEXT,
     related_id uuid,
     depth INTEGER,
-    tags jsonb,
+    labels jsonb,
```
moshloop, line 523: this should be tags

views/006_config_views.sql
```diff
@@ -368,9 +374,36 @@ FOR EACH ROW
...
+    ci.is_pushed, 
+    ci.last_scraped_time, 
+    ci.name, 
+    ci.parent_id, 
+    ci.path, 
+    ci.properties, 
+    ci.scraper_id, 
+    ci.source, 
+    ci.status, 
+    ci.labels,
```
moshloop, line 404: need tags here

views/006_config_views.sql
```diff
@@ -168,6 +169,11 @@ CREATE OR REPLACE VIEW config_tags AS
   SELECT d.key, d.value
   FROM configs JOIN json_each_text(tags::json) d ON true GROUP BY d.key, d.value ORDER BY key, value;
 
+-- config_labels
```
moshloop, line 172: we probably need a config_tags as well

adityathebe: the previous `config_tags` view also exists.

Issue #652, opened 2024-04-02
Changes API - Make catalog id optional


moshloop:
Don't use the recursive functions - just list changes

PR #653, opened 2024-04-03
support changes search without a catalog id

adityathebe:
resolves: https://github.com/flanksource/duty/issues/652


query/config_changes.go
```diff
@@ -108,12 +108,14 @@ func (t *CatalogChangesSearchRequest) SetDefaults() {
 }
 
 func (t *CatalogChangesSearchRequest) Validate() error {
-	if t.CatalogID == uuid.Nil {
-		return fmt.Errorf("catalog id is required")
-	}
+	if t.Recursive != "" {
```
moshloop, line 111: I think the recursive option should just default to downstream

PR #666, opened 2024-04-11
feat: remove type constraint from connection

adityathebe:


context/connection.go
```diff
@@ -77,7 +91,7 @@ func HydrateConnectionByURL(ctx Context, connectionString string) (*models.Conne
 			return models.ConnectionFromURL(*_url), nil
 		}
 
-		return nil, fmt.Errorf("invalid connection string: %q. Expected connection://<type>/<name> , uuid or URL", connectionString)
+		return nil, fmt.Errorf("invalid connection string: %q. Expected connetion string, uuid or URL", connectionString)
```
moshloop, line 94:
```suggestion
		return nil, fmt.Errorf("invalid connection string: %q. Expected connection string, uuid or URL", connectionString)
```

context/connection.go
```diff
@@ -117,29 +133,38 @@ func FindConnectionByURL(ctx Context, connectionString string) (*models.Connecti
...
 	if namespace == "" {
 		namespace = ctx.GetNamespace()
 	}
 
-	err := ctx.DB().Where("type = ? AND name = ? AND namespace = ?", connectionType, name, namespace).First(&connection).Error
-	if err != nil {
+	// NOTE: For backward compatibility reason we use the namespace as the connection type
+	// Before: connection://<type>/<name>
+	// Now: connection://<namespace>/<name.
+	if err := ctx.DB().Where("name = ? AND type = ?", name, namespace).
```
moshloop, line 160: Can we make this this the second option, and also log a warning when its found using this?

PR #671, opened 2024-04-16
feat: implement gauges, counters and histogram for ctx

yashmehrotra:
Fixes: https://github.com/flanksource/duty/issues/626


.github/workflows/test.yaml
```diff
@@ -31,7 +31,7 @@ jobs:
       - name: Install Go
         uses: actions/setup-go@0c52d547c9bc32b1aa3301fd7a9cb496313a4491 # v5.0.0
         with:
-          go-version: 1.20.x
+          go-version: 1.22.x
```
moshloop, line 34: do we not need to bump all the dependencies first ?

yashmehrotra: I built mission control and canary checker with this and there were no problems

yashmehrotra: I built mission control and canary checker with this and there were no problems

Issue #672, opened 2024-04-16
Add Config Server-Side Grouping API Endpoint


mainawycliffe:
API endpoint that will accept grouping parameters and return configs grouped by the specified parameters such as config types, tags, etc. Should also accept filter properties as well.

PR #678, opened 2024-04-17
feat: config summary new search query

adityathebe:
resolves: https://github.com/flanksource/duty/issues/672

### DIFF
--- a/query/config.go
+++ b/query/config.go
@@ -14,6 +14,7 @@ import (
 	"github.com/flanksource/duty/models"
 	"github.com/flanksource/duty/types"
 	"github.com/google/uuid"
+	"gorm.io/gorm"
 )
 
 type ConfigSummaryRequestChanges struct {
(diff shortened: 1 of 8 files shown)


query/config.go
```diff
@@ -4,15 +4,264 @@ import (
...
+	)
+
+	analysisGrouped := exclause.NewWith(
+		"analysis_grouped",
+		ctx.DB().Select(req.baseSelectClause("config_analysis.analysis_type", "COUNT(*) AS count")).
+			Model(&models.ConfigAnalysis{}).
+			Joins("LEFT JOIN config_items ON config_analysis.config_id = config_items.id").
+			Where(req.configDeleteClause()).
+			Where(req.filterClause(ctx.DB())).
+			Where("NOW() - config_analysis.first_observed <= ?", req.Analysis.sinceParsed).
```
moshloop, line 222: I think this would be on last observed

query/config.go
```diff
@@ -4,15 +4,264 @@ import (
...
+		t.Changes.Since = "2d"
+	}
+
+	if t.Analysis.Since == "" {
+		t.Analysis.Since = "2d"
+	}
+
+	if t.Cost == "" {
+		t.Cost = "30d"
+	}
```
moshloop, line 148: I don't think we should default these, If not specified we don't run those queries

Can we also add a health summary (healthy, unhealthy, unknown, warning, pending (!ready)

PR #693, opened 2024-04-22
feat: handle foreign key error during reconciliation

adityathebe:
The plan is that if the upstream gets any fk constraint error
- it does the insert/update one by one and keeps a list of all the items that failed due to fk constraint error
- the agent then marks the parent(s) of those failed items as `is_pushed=false`
- In the next round of reconciliation, the conflict will get resolved (or we could restart the round immediately after this round ends)

resolves: https://github.com/flanksource/duty/issues/689

### DIFF
--- a/models/config.go
+++ b/models/config.go
@@ -100,6 +100,14 @@ type ConfigItem struct {
 	DeleteReason    string               `json:"delete_reason,omitempty"`
 }
 
+func (t ConfigItem) UpdateParentsIsPushed(db *gorm.DB, items []DBTable) error {
+	parentIDs := lo.Map(items, func(item DBTable, _ int) string {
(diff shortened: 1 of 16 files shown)


models/config.go
```diff
@@ -71,7 +71,7 @@ var AllowedColumnFieldsInConfigs = []string{"config_class", "external_id"}
 
 // ConfigItem represents the config item database table
 type ConfigItem struct {
-	ID              uuid.UUID            `json:"id" faker:"uuid_hyphenated"`
+	ID              uuid.UUID            `json:"id" faker:"uuid_hyphenated" gorm:"default:generate_ulid()"`
```
moshloop, line 74: Why do we generate this in the db ?

Issue #690, opened 2024-04-22
Resource selectors - remove validation of label names


moshloop:
error running scraper[Infoslips]: failed to save results: failed to update db: failed to get relationship results from relationship selectors: failed to find config items by relationship selector: error: code=invalid message=failed to parse label selector: unable to parse requirement: values[0][aws/iam-role]: Invalid value: "arn:aws:iam::804104647664:role/AmazonEKSVPCCNIRole": a valid label must be an empty string or consist of alphanumeric characters, '-', '_' or '.', and must start and end with an alphanumeric character (e.g. 'MyValue', or 'my_value', or '12345', regex used for validation is '(([A-Za-z0-9][-A-Za-z0-9_.]*)?[A-Za-z0-9])?')

PR #699, opened 2024-04-24
feat: custom label selector parser with no validation

adityathebe:
resolves: https://github.com/flanksource/duty/issues/690


pkg/kube/labels/selector_test.go
```diff
@@ -0,0 +1,1055 @@
...
+		"",
+		"x!=a,y=b",
+		"x=",
+		"x= ",
+		"x=,z= ",
+		"x= ,z= ",
+		"!x",
+		"x>1",
+		"x>1,z<5",
+		"aws/iam-role=arn:aws:iam::12345678910:role/AmazonEKSVPCCNIRole",
```
adityathebe, line 48: test case here (for visibility)

PR #713, opened 2024-04-29
fix: unique constraints

adityathebe:


adityathebe:
@moshloop `people` table didn't have a `deleted_at` column. I've added it here.

PR #717, opened 2024-05-01
fix: config summary

adityathebe:
This was an edge case bug where the following query would generate an invalid sql query.

```sh
curl -s -X POST -u 'admin@local:admin' http://localhost:8080/catalog/summary \
--json '{
  "groupBy": ["type", "namespace", "region"],
  "deleted": false,
  "changes": { "since": "" },
  "analysis": { "since": "30d" },
  "cost": "30d"
}'
```

What's weird is that it works fine if you group by `1` or `2` columns. As soon as you group by 3 columns (doesn't matter which), this bug happens. If you group by `4` columns then it works fine and then fails again if you group by `5` columns. _(probably fails if you have odd columns but I didn't test further)_

The bug particularly affects the `aggregated_health_count` `WITH` clause and in particular the select clause of that query. Here's the generated invalid query

```sql
WITH "aggregated_health_count" AS (
  SELECT
    TYPE,
    namespace,
    region,
    json_object_agg(analysis_type, count) :: jsonb AS analysis -- somehow the select clause of `aggregated_analysis_count` WITH clause gets used here
  FROM
    "health_grouped"
  GROUP BY
    TYPE,
    namespace,
    region
)
```

### DIFF
--- a/query/config.go
+++ b/query/config.go
@@ -15,6 +15,7 @@ import (
 	"github.com/flanksource/duty/types"
 	"github.com/google/uuid"
 	"gorm.io/gorm"
+	"gorm.io/gorm/clause"
 )
 
 type ConfigSummaryRequestChanges struct {
@@ -66,7 +67,7 @@ func (t *ConfigSummaryRequest) analysisJoin() string {
 	for _, g := range t.GroupBy {
 		switch g {
 		case "type":
-			clauses = append(clauses, "aggregated_analysis_count .type = config_items.type")
+			clauses = append(clauses, "aggregated_analysis_count.type = config_items.type")
 		default:
 			clauses = append(clauses, fmt.Sprintf("aggregated_analysis_count.%s = config_items.tags->>'%s'", g, g))
 		}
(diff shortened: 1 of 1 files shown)


query/config.go
```diff
@@ -90,8 +91,11 @@ func (t *ConfigSummaryRequest) changesJoin() string {
 	return output + strings.Join(clauses, " AND ")
 }
 
-func (t *ConfigSummaryRequest) plainSelectClause(appendSelect ...string) []string {
-	return append(t.GroupBy, appendSelect...)
```
adityathebe, line 94: I'm pretty sure the bug was happening here but not 100%.

What was happening is that the select columns generated by this function were internally being mutated by gorm.

Example: the first call returns columns 
`first := ["type", "namespace", "region", "jsonb_object_agg(health_grouped.health, count)::jsonb AS health"]`
and the second call returns 
`second := ["type", "namespace", "region", "json_object_agg(analysis_type, count)::jsonb AS analysis"]`

after the second call, `first` becomes idential to `second`

Issue #708, opened 2024-04-26
catalog change API should return created by


mainawycliffe:
- User ID
- Name
- Avatar

moshloop:
+ handle the case of external created by

PR #728, opened 2024-05-02
chore: include created by details in catalog change api

yashmehrotra:
Fixes https://github.com/flanksource/duty/issues/708

### DIFF
--- a/query/config_changes.go
+++ b/query/config_changes.go
@@ -162,8 +162,6 @@ type ConfigChangeRow struct {
 	ConfigType        string     `gorm:"column:type" json:"type,omitempty"`
 	CreatedBy         *uuid.UUID `gorm:"column:created_by" json:"created_by,omitempty"`
 	ExternalCreatedBy string     `gorm:"column:external_created_by" json:"external_created_by,omitempty"`
-	AuthorName        string     `json:"author_name,omitempty"`
-	AuthorAvatar      string     `json:"author_avatar,omitempty"`
 }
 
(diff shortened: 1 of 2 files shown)


query/config_changes.go
```diff
@@ -150,16 +150,20 @@ func (t *CatalogChangesSearchRequest) Validate() error {
...
+	Severity          string     `gorm:"column:severity" json:"severity"  faker:"oneof: critical, high, medium, low, info"`
+	Source            string     `gorm:"column:source" json:"source"`
+	Summary           string     `gorm:"column:summary;default:null" json:"summary,omitempty"`
+	CreatedAt         *time.Time `gorm:"column:created_at" json:"created_at"`
+	ConfigName        string     `gorm:"column:name" json:"name,omitempty"`
+	ConfigType        string     `gorm:"column:type" json:"type,omitempty"`
+	CreatedBy         *uuid.UUID `gorm:"column:created_by" json:"created_by,omitempty"`
+	ExternalCreatedBy string     `gorm:"column:external_created_by" json:"external_created_by,omitempty"`
+	AuthorName        string     `json:"author_name,omitempty"`
+	AuthorAvatar      string     `json:"author_avatar,omitempty"`
```
moshloop, line 166: I don't think we need to include these - The vast majority of changes will not be linked to a person and we can do the lookup and cache of person -> name, avatar client side

yashmehrotra: So just return source, external_created_by and created_by and let frontend decide ?

Issue #727, opened 2024-05-01
Cleanup stale jobs on startup


moshloop:
When a controller panics running jobs do not get cleaned up on restart

e.g. If there are running jobs when it initialises (on startup) mark them as cancelled

PR #736, opened 2024-05-03
feat: clear out stale running job histories for singleton jobs

adityathebe:
resolves: https://github.com/flanksource/duty/issues/727

1. The stale running job histories of singleton jobs will now be cleared by the status ring itself.
2. For stale running job histories of deleted resources _(i.e. the job doesn't run anymore)_, the cleanup job in mission control will call `CleanupStaleHistoryJob` with a smaller stale age _(probably 4 hours)_ just for running statuses.

### DIFF
--- a/job/cleanup.go
+++ b/job/cleanup.go
@@ -8,32 +8,15 @@ import (
 	"github.com/flanksource/duty/models"
 )
 
-func CleanupStaleHistoryJob(ctx context.Context, age time.Duration, name, resourceID string) *Job {
-	return &Job{
-		Context:    ctx,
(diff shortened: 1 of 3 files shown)


job/cleanup.go
```diff
@@ -8,33 +8,16 @@ import (
 	"github.com/flanksource/duty/models"
 )
 
-func CleanupStaleHistoryJob(ctx context.Context, age time.Duration, name, resourceID string) *Job {
```
adityathebe, line 11: Will define this job in mission-control itself.

Issue #739, opened 2024-05-05
Set role based default timeouts


moshloop:
On migrate attempt to set default timeouts for 
db.postgrest.timeout=60s
db.timeout=60m

```
ALTER ROLE postgrest_api SET statement_timeout = '120s';
ALTER ROLE connection user name SET statement_timeout = '120s';
```

if either fails log a warning and continue

PR #748, opened 2024-05-07
feat: set statement timeouts on roles

adityathebe:
resolves: https://github.com/flanksource/duty/issues/739#event-12710898365

### DIFF
--- a/functions/postgrest.sql
+++ b/functions/postgrest.sql
@@ -6,8 +6,10 @@ BEGIN
         ALTER DEFAULT PRIVILEGES IN SCHEMA public GRANT SELECT, UPDATE, DELETE, INSERT ON TABLES TO postgrest_api;
         GRANT USAGE, SELECT ON ALL SEQUENCES IN SCHEMA public TO postgrest_api;
     END IF;
+    ALTER ROLE postgrest_api SET statement_timeout = '60s';
 
     IF NOT EXISTS (SELECT FROM pg_catalog.pg_roles WHERE rolname = 'postgrest_anon') THEN
         CREATE ROLE postgrest_anon;
     END IF;
+    ALTER ROLE postgrest_anon SET statement_timeout = '60s';
 END $$;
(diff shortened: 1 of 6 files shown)


migrate/migrate.go
```diff
@@ -76,16 +89,12 @@ func RunMigrations(pool *sql.DB, connection string, opts MigrateOptions) error {
...
-	parsedConn, err := url.Parse(connection)
-	if err != nil {
-		return err
-	}
-	user := parsedConn.User.Username()
-	if user == "" {
-		return fmt.Errorf("malformed connection string, got empty username: %s", parsedConn.Redacted())
-	}
+func setTimeouts(pool *sql.DB, user string) error {
+	_, err := pool.Exec(fmt.Sprintf(`ALTER ROLE %s SET statement_timeout = '60min'`, user))
```
moshloop, line 93: The timeout needs to be configurable

functions/postgrest.sql
```diff
@@ -6,8 +6,10 @@ BEGIN
         ALTER DEFAULT PRIVILEGES IN SCHEMA public GRANT SELECT, UPDATE, DELETE, INSERT ON TABLES TO postgrest_api;
         GRANT USAGE, SELECT ON ALL SEQUENCES IN SCHEMA public TO postgrest_api;
     END IF;
+    ALTER ROLE postgrest_api SET statement_timeout = '60s';
```
moshloop, line 9: This needs to be configurable, and be updated even after the role has been created

migrate/migrate.go
```diff
@@ -51,14 +72,25 @@ func RunMigrations(pool *sql.DB, connection string, opts MigrateOptions) error {
...
+	}
+
 	logger.Tracef("Granting roles to current user")
 	// Grant postgrest roles in ./functions/postgrest.sql to the current user
-	if err := grantPostgrestRolesToCurrentUser(pool, connection); err != nil {
+	if err := grantPostgrestRolesToCurrentUser(pool, user); err != nil {
 		return fmt.Errorf("failed to grant postgrest roles: %w", err)
 	}
 
+	setStatementTimeouts(ctx, pool, user, opts)
```
moshloop, line 90: I think the statement timeous should come after the migrations have run, especially as the timeout value could come from the properties table

moshloop: We also then wouldn't need to change the signatures of Init/Migrate

PR #750, opened 2024-05-07
feat: add status filter in catalog api

yashmehrotra:
Fixes: #731

### DIFF
--- a/query/config.go
+++ b/query/config.go
@@ -37,6 +37,7 @@ type ConfigSummaryRequest struct {
 	GroupBy  []string                     `json:"groupBy"`
 	Tags     []string                     `json:"tags"`
 	Status   string                       `json:"status"`
+	Health   string                       `json:"health"`
 }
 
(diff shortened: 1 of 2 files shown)


query/config.go
```diff
@@ -197,14 +198,22 @@ func (t *ConfigSummaryRequest) Parse() error {
...
+func (t ConfigSummaryRequest) configDeleteClause() string {
 	if !t.Deleted {
 		return "config_items.deleted_at IS NULL"
 	}
 
 	return ""
 }
 
+func (t ConfigSummaryRequest) statusClause(q *gorm.DB) *gorm.DB {
+	if t.Status != "" {
```
moshloop, line 210: Can we use https://github.com/flanksource/duty/blob/main/query/commons.go#L13 ? 

I think we also need a health filter

That function can probably also be optimized to return equals / not equals instead of in / not in for single values

PR #753, opened 2024-05-08
chore: remove push_queue_summary view

adityathebe:


views/007_events.sql
```diff
@@ -217,22 +217,6 @@ AFTER UPDATE ON components
 FOR EACH ROW
 EXECUTE PROCEDURE insert_component_health_updates_in_event_queue();
 
-CREATE OR REPLACE VIEW push_queue_summary AS
```
moshloop, line 220: Shouldn't this then be a view with a count of pending items per table ?

PR #761, opened 2024-05-10
feat: hard/soft on config_relationships

adityathebe:


views/006_config_views.sql
```diff
@@ -640,3 +649,22 @@ CREATE OR REPLACE VIEW config_detail AS
         WHERE start_time > NOW() - interval '30 days' 
         GROUP BY config_id) as playbook_runs
       ON ci.id = playbook_runs.config_id;
+
+-- config parent to config_relationship trigger
+CREATE OR REPLACE FUNCTION insert_parent_to_config_relationship()
+RETURNS TRIGGER AS $$
+BEGIN
+  IF NEW.parent_id IS NOT NULL THEN
```
moshloop, line 657: Can we check if the parent_id has changed and only then insert ?

views/006_config_views.sql
```diff
@@ -654,3 +663,53 @@ CREATE OR REPLACE VIEW config_detail AS
...
+END
+$$ LANGUAGE plpgsql;
+
+CREATE OR REPLACE TRIGGER fn_insert_parent_to_config_relationship
+AFTER INSERT OR UPDATE ON config_items
+FOR EACH ROW
+EXECUTE PROCEDURE insert_parent_to_config_relationship();
+
+-- Add the hard relationship for records existing before the trigger was added
+INSERT INTO config_relationships (config_id, related_id, relation)
```
moshloop, line 702: I think this needs to be moved to an individual file, and not run every time config_views is updated

adityathebe: good point 👍

PR #785, opened 2024-05-21
feat: push job history to upstream

yashmehrotra:
Fixes: https://github.com/flanksource/mission-control/issues/701

### DIFF
--- a/schema/system.hcl
+++ b/schema/system.hcl
@@ -211,4 +211,8 @@ table "job_history" {
   primary_key {
     columns = [column.id]
   }
+  index "job_history_is_pushed_idx" {
+    columns = [column.is_pushed]
+    where   = "is_pushed IS FALSE"
+  }
 }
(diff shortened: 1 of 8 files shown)


schema/system.hcl
```diff
@@ -198,6 +203,11 @@ table "job_history" {
     type    = timestamptz
     default = sql("now()")
   }
+  column "is_pushed" {
```
moshloop, line 206: I think we need an is_pushed index

yashmehrotra: Added 👍🏽

schema/system.hcl
```diff
@@ -198,7 +203,16 @@ table "job_history" {
...
+    null    = false
+    default = false
+    type    = bool
+  }
   primary_key {
     columns = [column.id]
   }
+  index "job_history_is_pushed_idx" {
+    columns = [column.is_pushed]
+    where   = "is_pushed IS FALSE"
```
moshloop, line 216: This should match the where clause above

yashmehrotra: My bad, fixed it

PR #790, opened 2024-05-22
feat: in config & component health update triggers, add additional metadata to the event properties

adityathebe:


views/007_events.sql
```diff
@@ -174,7 +174,7 @@ BEGIN
     event_name := CONCAT('config.', COALESCE(NULLIF(NEW.health, ''), 'unknown'));
 
     INSERT INTO event_queue(name, properties)
-    VALUES (event_name, jsonb_build_object('id', NEW.id))
+    VALUES (event_name, jsonb_build_object('id', NEW.id, 'status', NEW.status, 'description', NEW.description))
```
moshloop, line 177: I think we need health as well

PR #841, opened 2024-06-14
fix: retry on deadlock

adityathebe:
resolves: https://github.com/flanksource/config-db/issues/637

### DIFF
--- a/upstream/jobs.go
+++ b/upstream/jobs.go
@@ -1,13 +1,17 @@
 package upstream
 
 import (
+	gocontext "context"
 	"encoding/json"
 	"fmt"
+	"time"
 
+	"github.com/flanksource/duty"
 	"github.com/flanksource/duty/api"
 	"github.com/flanksource/duty/context"
 	"github.com/flanksource/duty/models"
 	"github.com/samber/lo"
+	"github.com/sethvargo/go-retry"
 	"gorm.io/gorm"
 )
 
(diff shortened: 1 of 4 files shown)


upstream/jobs.go
```diff
@@ -134,15 +138,33 @@ func reconcileTable(ctx context.Context, config UpstreamConfig, table pushableTa
...
 
-		if c, ok := table.(customIsPushedUpdater); ok {
-			if err := c.UpdateIsPushed(ctx.DB(), items); err != nil {
-				return 0, fmt.Errorf("failed to update is_pushed for %s: %w", table.TableName(), err)
-			}
-		} else {
-			ids := lo.Map(items, func(a models.DBTable, _ int) string { return a.PK() })
-			if err := ctx.DB().Model(table).Where("id IN ?", ids).Update("is_pushed", true).Error; err != nil {
-				return 0, fmt.Errorf("failed to update is_pushed on %s: %w", table.TableName(), err)
+		backoff := retry.WithMaxRetries(3, retry.NewConstant(time.Second))
```
moshloop, line 141:
```suggestion
       backoff = time.Second
       if len(items) > 500:
            backoff = 3 * time.Second
       elseif len(items) > 100:
               backoff = 2 * time.Second
        
		backoff := retry.WithMaxRetries(3, retry.NewExponential(backoff)).WithJitter(0.2)
```

We should probably also be updating in batches

upstream/jobs.go
```diff
@@ -134,14 +138,35 @@ func reconcileTable(ctx context.Context, config UpstreamConfig, table pushableTa
...
 
-		if c, ok := table.(customIsPushedUpdater); ok {
-			if err := c.UpdateIsPushed(ctx.DB(), items); err != nil {
-				return 0, fmt.Errorf("failed to update is_pushed for %s: %w", table.TableName(), err)
-			}
-		} else {
-			ids := lo.Map(items, func(a models.DBTable, _ int) string { return a.PK() })
-			if err := ctx.DB().Model(table).Where("id IN ?", ids).Update("is_pushed", true).Error; err != nil {
-				return 0, fmt.Errorf("failed to update is_pushed on %s: %w", table.TableName(), err)
+		batchSize := 50
```
moshloop, line 141:
```suggestion
		batchSize := ctx.Properties().int("pushed.batch.size", 200)
```

PR #871, opened 2024-06-28
feat: cleanup agent job history

yashmehrotra:


db/utils.go
```diff
@@ -0,0 +1,48 @@
+package db
```
yashmehrotra, line 1: Moved few funcs here to avoid import cycle errors

PR #873, opened 2024-06-28
chore: create checks_by_config and check_summary_by_config views

mainawycliffe:


views/006_config_views.sql
```diff
@@ -765,3 +765,34 @@ CREATE or REPLACE VIEW config_statuses AS
 -- Cleanup old trigger that used to add 'push_queue.create' events
 -- for any changes on a selected list of tables.
 DROP FUNCTION IF EXISTS push_changes_to_event_queue CASCADE;
+
+-- checks_by_config
```
moshloop, line 769: I think model this after: https://github.com/flanksource/duty/blob/main/views/013_check_summary.sql#L128

PR #885, opened 2024-07-03
feat: remove label key validation

adityathebe:
This is required to support looking up labels with key like `aws:cloudformation:stack-id`

### DIFF
--- a/pkg/kube/labels/selector_test.go
+++ b/pkg/kube/labels/selector_test.go
@@ -460,17 +460,17 @@ func TestRequirementConstructor(t *testing.T) {
 				},
 			},
 		},
-		{
-			Key: strings.Repeat("a", 254), //breaks DNS rule that len(key) <= 253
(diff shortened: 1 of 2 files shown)


pkg/kube/labels/selector_test.go
```diff
@@ -460,17 +460,17 @@ func TestRequirementConstructor(t *testing.T) {
...
-			Key: strings.Repeat("a", 254), //breaks DNS rule that len(key) <= 253
-			Op:  selection.Exists,
-			WantErr: field.ErrorList{
-				&field.Error{
-					Type:     field.ErrorTypeInvalid,
-					Field:    "key",
-					BadValue: strings.Repeat("a", 254),
-				},
-			},
-		},
```
adityathebe, line 473: we don't have any restriction when storing labels, so we probably shouldn't have it when quering them.

PR #889, opened 2024-07-08
Config generator

moshloop:


tests/config_generator.go
```diff
@@ -0,0 +1,144 @@
...
+
+func (generator *ConfigGenerator) GenerateConfigItem(configType string, parent *models.ConfigItem) models.ConfigItem {
+	changes := []models.ConfigChange{}
+	analysis := []models.ConfigAnalysis{}
+	generator.count++
+	name := fmt.Sprintf("%s-%d", strings.Split(configType, "::")[1], generator.count)
+
+	item := models.ConfigItem{
+		ID:        uuid.New(),
+		DeletedAt: generator.GenerateDeleted(),
```
adityathebe, line 63: @moshloop I think we want this to be deterministic to be used in tests.

moshloop: Ok - So maybe setup a real structure so that the deployment have 1 active replicase with a pod, and 3 replicasets scaled to zero, each with deleted pods

tests/config_relationship_test.go
```diff
@@ -493,3 +493,72 @@ var _ = ginkgo.Describe("config relationship deletion test", func() {
...
+
+	ginkgo.AfterAll(func() {
+		err := generator.Destroy(DefaultContext.DB())
+		Expect(err).To(BeNil())
+
+		assertConfigCount("", DeleteFilterNone, 0)
+	})
+
+	ginkgo.It("should fetch level 1", func() {
+		cluster := generator.Generated.ConfigByTypes("Kubernetes::Cluster")[0]
```
moshloop, line 522: Can we add some tests from deployment

PR #897, opened 2024-07-11
feat: http connection

yashmehrotra:


connection/http.go
```diff
@@ -0,0 +1,161 @@
...
+	h.TLS.Cert.ValueStatic, err = ctx.GetEnvValueFromCache(h.TLS.Cert, namespace)
+	if err != nil {
+		return h, err
+	}
+	return h, nil
+}
+
+// CreateHTTPClient requires a hydrated connection
+func CreateHTTPClient(ctx ConnectionContext, conn HTTPConnection) (*http.Client, error) {
+	client := http.NewClient()
```
moshloop, line 84: Why not just hydrate here ?

yashmehrotra: Won't have access to the namespace here

PR #905, opened 2024-07-11
feat(resource selector): add scope

adityathebe:


relationship_selector.go
```diff
@@ -57,10 +57,16 @@ type RelationshipSelector struct {
 	Type       string            `json:"type,omitempty"`
 	Agent      string            `json:"agent,omitempty"`
 	Labels     map[string]string `json:"labels,omitempty"`
+
+	// Scope is the id parent of the resource to select.
+	// Example: For config items, the scope is the scraper id
+	// - for checks, it's canaries and
+	// - for components, it's topology.
+	Scope string `json:"scope,omitempty"`
```
adityathebe, line 65: Only supports id and not name as we don't have unique constraint on the name.

Issue #928, opened 2024-07-31
Resource Selectors - Add search field


moshloop:
Startswith query on name label and tag values

PR #929, opened 2024-07-31
feat: Prefix search on query selector

adityathebe:
resolves: https://github.com/flanksource/duty/issues/928


types/resource_selector.go
```diff
@@ -34,7 +34,10 @@ type ResourceSelector struct {
 	//  'max-age=X' (cache for X duration)
 	Cache string `yaml:"cache,omitempty" json:"cache,omitempty"`
 
-	IncludeDeleted bool `yaml:"-" json:"-"`
```
adityathebe, line 37: not sure if this was intentionally kept hidden from json/yaml.

PR #953, opened 2024-08-15
feat: jitter on cron jobs

adityathebe:


job/job.go
```diff
@@ -489,20 +497,44 @@ func (j *Job) AddToScheduler(cronRunner *cron.Cron) error {
...
-	j.entryID = &entryID
+
+	if j.Jitter == 0 {
+		entryID, err := cronRunner.AddJob(schedule, j)
+		if err != nil {
+			return fmt.Errorf("[%s] failed to schedule job: %s", j.Label(), err)
+		}
+		j.entryID = &entryID
+	} else {
+		delaydJob := cron.FuncJob(func() {
```
moshloop, line 515: I think the jitter should be once off at the start, i.e. delay the scheduling of the job by random(min(duration till next runtime, 1h))

adityathebe:
@moshloop I'm leaving off delaying the initial scheduling even though we discussed we would add a 100% jitter to it.

The problem is that if we delay the scheduling by running `entryID, err := cronRunner.AddJob(schedule, delayedJob)` in a goroutine, the `AddToScheduler` func returns with no error but the job actually hasn't been scheduled. i.e. job.entryID = nil.

If the job needs to be rescheduled (scraper CRD updated) when it's still waiting for first schedule, then it's not possible to unschedule is as we won't have the `entryID`.

I think let's try with per iteration jitter for now and see how it goes?

job/job.go
```diff
@@ -3,7 +3,9 @@ package job
 import (
 	"container/ring"
 	gocontext "context"
+	"crypto/rand"
```
moshloop, line 6: We don't want to use crypto/rand which can block if there is not enough entropy - use math/rand

job/job.go
```diff
@@ -26,6 +28,11 @@ const (
 	ResourceTypeUpstream      = "upstream"
 )
 
+const (
+	// iterationJitterPercent sets the maximum percent by which to jitter each subsequent invocation of a periodic job
+	iterationJitterPercent = 10
```
moshloop, line 33: I think we will need a max percent, and max.duration (15m)

job/job.go
```diff
@@ -511,20 +517,39 @@ func (j *Job) AddToScheduler(cronRunner *cron.Cron) error {
...
+	parsedSchedule, err := cron.ParseStandard(j.Schedule)
+	if err != nil {
+		return fmt.Errorf("failed to parse schedule: %s", err)
+	}
+	interval := time.Until(parsedSchedule.Next(time.Now()))
+
+	delayedJob := cron.FuncJob(func() {
+		delayPercent, _ := rand.Int(rand.Reader, big.NewInt(int64(iterationJitterPercent)))
+		jitterDuration := time.Duration((int64(interval) * delayPercent.Int64()) / 100)
+		j.Context.Logger.Named(j.GetResourcedName()).V(2).Infof("jitter (%d%%): %s", delayPercent, jitterDuration)
```
moshloop, line 534:
```suggestion
		j.Context.V(4).Infof("jitter (%d%%): %s", delayPercent, jitterDuration)
```

job/job.go
```diff
@@ -511,20 +518,43 @@ func (j *Job) AddToScheduler(cronRunner *cron.Cron) error {
...
+	if err != nil {
+		return fmt.Errorf("failed to parse schedule: %s", err)
+	}
+	interval := time.Until(parsedSchedule.Next(time.Now()))
+
+	delayedJob := cron.FuncJob(func() {
+		delayPercent := rand.Intn(iterationJitterPercent)
+		jitterDuration := time.Duration((int64(interval) * int64(delayPercent)) / 100)
+		if jitterDuration > maxJitterDuration {
+			jitterDuration = maxJitterDuration
```
moshloop, line 536: I think this will have fixed 15m jitters, so if interval * delayPercent then interval = 15m

PR #960, opened 2024-08-16
feat: embedded DB setup

adityathebe:


start.go
```diff
@@ -19,12 +25,16 @@ func BindPFlags(flags *pflag.FlagSet) {
...
+	flags.BoolVar(&DefaultConfig.RunMigrations, "db-migrations", false, "Run database migrations")
 	flags.BoolVar(&DefaultConfig.SkipMigrations, "skip-migrations", false, "Skip database migrations")
 	flags.BoolVar(&DefaultConfig.Postgrest.Disable, "disable-postgrest", false, "Disable PostgREST. Deprecated (Use --postgrest-uri '' to disable PostgREST)")
 	flags.StringVar(&DefaultConfig.Postgrest.DBAnonRole, "postgrest-anon-role", "postgrest-api", "PostgREST anonymous role")
 	flags.IntVar(&DefaultConfig.Postgrest.MaxRows, "postgrest-max-rows", 2000, "A hard limit to the number of rows PostgREST will fetch")
 	flags.StringVar(&DefaultConfig.LogLevel, "db-log-level", "error", "Set gorm logging level. trace, debug & info")
 	flags.BoolVar(&DefaultConfig.DisableKubernetes, "disable-kubernetes", false, "Disable Kubernetes integration")
+	flags.BoolVar(&DefaultConfig.Metrics, "db-metrics", false, "Expose db metrics")
+
+	_ = flags.MarkDeprecated("skip-migrations", "migrations are skipped by default. Use --db-migrations to run migrations.")
```
moshloop, line 37: I don't think we should skip by default - it means the default arguments will produce errors on an empty db - for mission-control chart we can't run migrations from multiple pods/versions and then for skip when bundling

PR #978, opened 2024-08-22
fix: topology view

adityathebe:
resolves: failed to query rest of the children: ERROR: column "health_expr" does not exist (SQLSTATE 42703)

`health_expr` exists however the topology view needs to be recreated.

A small diff on the migration file should trigger the run.

### DIFF
--- a/hack/migrate/main.go
+++ b/hack/migrate/main.go
@@ -13,6 +13,7 @@ var migrate = &cobra.Command{
 	Run: func(cmd *cobra.Command, args []string) {
 		if err := duty.Migrate(api.Config{
 			ConnectionString: connection,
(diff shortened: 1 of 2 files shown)


hack/migrate/main.go
```diff
@@ -13,6 +13,7 @@ var migrate = &cobra.Command{
 	Run: func(cmd *cobra.Command, args []string) {
 		if err := duty.Migrate(api.Config{
 			ConnectionString: connection,
+			Postgrest:        api.DefaultConfig.Postgrest,
```
adityathebe, line 16: resolves: `ERR+1 failed to grant postgrest roles: ERROR: syntax error at or near "TO" (SQLSTATE 42601)`

PR #996, opened 2024-08-26
feat: child job histories

adityathebe:


job/job.go
```diff
@@ -228,15 +225,37 @@ func (j *JobRuntime) ID() string {
 func (j *JobRuntime) start() {
 	j.Tracef("starting")
 	j.Context.Counter("job_started", "name", j.Job.Name, "id", j.Job.ResourceID, "resource", j.Job.ResourceType).Add(1)
-	j.History = models.NewJobHistory(j.Logger, j.Job.Name, "", "").Start()
+
+	if j.Parent != "" {
+		query := j.Context.DB().Where("name = ?", j.Parent)
```
moshloop, line 230: I don't think we can do this - this will put a log of reads onto the table

moshloop: It might be better for us to just implement a ring for job histories that are pushed

Issue #1002, opened 2024-08-29
Topology Health: Rewrite logic


yashmehrotra:
Unhealthy if:
- any child is unhealthy
- any check is unhealthy
- component itself is unhealthy

PR #1004, opened 2024-08-30
chore: revamp component health logic

yashmehrotra:
Fixes: #1002

### DIFF
--- a/models/components.go
+++ b/models/components.go
@@ -192,21 +192,20 @@ func (c Component) GetHealth() (Health, error) {
 	}
 
 	// When HealthExpr is not defined, we take worse of checks, children and the component itself
-	if h := lo.FromPtr(c.Health); h == HealthUnhealthy {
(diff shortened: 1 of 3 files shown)


models/components.go
```diff
@@ -188,24 +188,25 @@ func (c Component) GetHealth() (string, error) {
 			return "", fmt.Errorf("failed to evaluate health expression %s: %v", c.HealthExpr, err)
 		}
 
-		return out, nil
+		return Health(out), nil
 	}
 
-	if h := lo.FromPtr(c.Health); h != "" {
-		return string(h), nil
+	// When HealthExpr is not defined, we take worse of checks, children and the component itself
```
moshloop, line 194: This ignores warning status - Should probably make health comparable and then sort it pick the max

yashmehrotra: 👍🏽

Issue #998, opened 2024-08-27
For Resource Selector - When Searching - Return a Limited Number of Results


mainawycliffe:
When the list of items is high, it causes the UI to freeze and the user can refine the search

moshloop:
Add limit to search request body with default of 100

PR #1013, opened 2024-09-04
feat: limit on resource selector search

adityathebe:
resolves: https://github.com/flanksource/duty/issues/998


query/resource_selector.go
```diff
@@ -23,6 +23,8 @@ import (
 )
 
 type SearchResourcesRequest struct {
+	PageSize int `json:"page_size"`
```
moshloop, line 26: I think this should be limit as we are not implementing paging

PR #1014, opened 2024-09-05
feat: config.updated events

adityathebe:


views/006_config_views.sql
```diff
@@ -383,6 +381,33 @@ AFTER INSERT OR UPDATE ON config_items
 FOR EACH ROW
   EXECUTE FUNCTION insert_config_create_update_delete_in_event_queue();
 
+---
+CREATE OR REPLACE FUNCTION insert_config_changes_updates_in_event_queue()
+RETURNS TRIGGER AS
+$$
+DECLARE
+  event_name TEXT := 'config.changes';
```
moshloop, line 389:
```suggestion
  event_name TEXT := 'config.change';
```

adityathebe: I think should be `config.changed`

Issue #1006, opened 2024-08-30
Catalog Summary - Filter by Label - Support Include/Exclude For Labels


mainawycliffe:
Multiple Label Include returns empty

```
{
    "groupBy": [
        "type"
    ],
    "deleted": false,
    "filter": {
        "account": "765618022540",
        "alertmanager": "kube-prometheus-stack-alertmanager"
    },
    "changes": {
        "since": "30d"
    },
    "analysis": {
        "since": "30d"
    },
    "cost": "30d"
}
```

![image](https://github.com/user-attachments/assets/fd299d9f-d3b9-42ab-8dd9-95d43f66469b)

Exclude doesn't work at all:

```
{
    "groupBy": [
        "type"
    ],
    "deleted": false,
    "filter": {
        "account": "!765618022540"
    },
    "changes": {
        "since": "30d"
    },
    "analysis": {
        "since": "30d"
    },
    "cost": "30d"
}
```

![image](https://github.com/user-attachments/assets/90df2c0b-678f-4a08-a529-8e50c3a94f8b)

mainawycliffe:
Related to https://github.com/flanksource/flanksource-ui/pull/2188

PR #1028, opened 2024-09-09
feat: support include/exclude for labels filter in config summary

adityathebe:
resolves: https://github.com/flanksource/duty/issues/1006

### DIFF
--- a/query/config.go
+++ b/query/config.go
@@ -218,29 +218,39 @@ func (t ConfigSummaryRequest) healthClause() []clause.Expression {
 }
 
 func (t *ConfigSummaryRequest) filterClause(q *gorm.DB) *gorm.DB {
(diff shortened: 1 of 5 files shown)


query/config.go
```diff
@@ -218,8 +218,22 @@ func (t ConfigSummaryRequest) healthClause() []clause.Expression {
 }
 
 func (t *ConfigSummaryRequest) filterClause(q *gorm.DB) *gorm.DB {
+	var orClause *gorm.DB
 	for k, v := range t.Filter {
-		q = q.Where("config_items.labels @> ?", types.JSONStringMap{k: v})
+		if strings.HasPrefix(v, "!") {
```
moshloop, line 223: Can we not use the same code as the middleware - i.e. to support multiple values

PR #1030, opened 2024-09-09
feat: notification silence

adityathebe:


schema/notifications.hcl
```diff
@@ -165,3 +165,65 @@ table "notification_send_history" {
...
 }
+
+table "notification_silences" {
+  schema = schema.public
+  column "id" {
+    null    = false
+    type    = uuid
+    default = sql("generate_ulid()")
+  }
+  column "namespace" {
```
moshloop, line 176: Why do we need this ?

adityathebe: Don't we need a namespace column for all our CRDs?

PR #1061, opened 2024-09-17
feat: notification delay & event_queue delay

adityathebe:


postq/event.go
```diff
@@ -35,6 +35,7 @@ func fetchEvents(ctx context.Context, tx *gorm.DB, watchEvents []string, batchSi
 		WHERE id IN (
 			SELECT id FROM event_queue
 			WHERE
+				(delay IS NULL OR created_at + (delay * INTERVAL '1 second' / 1000000000)  <= NOW()) AND
```
moshloop, line 38: Might need to update the indexes for this

PR #1063, opened 2024-09-17
fix: add health to component topology select clause

yashmehrotra:
Fixes: https://github.com/flanksource/canary-checker/issues/2143


yashmehrotra:
@moshloop Self merging this for now

PR #1084, opened 2024-09-23
chore: select related components via config

yashmehrotra:
Adding it into resourceSelector.Search was tricky due to existing flow and handling args


query/resource_selector.go
```diff
@@ -225,18 +226,41 @@ func SetResourceSelectorClause(ctx context.Context, resourceSelector types.Resou
...
-		if resourceSelector.Name == "" {
-			prefixQueries = append(prefixQueries, ctx.DB().Where("name ILIKE ?", resourceSelector.Search+"%"))
-		}
-		if resourceSelector.TagSelector == "" && table == "config_items" {
-			prefixQueries = append(prefixQueries, ctx.DB().Where("EXISTS (SELECT 1 FROM jsonb_each_text(tags) WHERE value ILIKE ?)", resourceSelector.Search+"%"))
-		}
-		if resourceSelector.LabelSelector == "" {
-			prefixQueries = append(prefixQueries, ctx.DB().Where("EXISTS (SELECT 1 FROM jsonb_each_text(labels) WHERE value ILIKE ?)", resourceSelector.Search+"%"))
+		if strings.HasPrefix(resourceSelector.Search, "component_config_traverse") && table == "components" {
+			// search: component_config_traverse=72143d48-da4a-477f-bac1-1e9decf188a6,outgoing,Kubernetes::Pod,Kubernetes::Node
```
moshloop, line 230: Can we use this parser to start with: https://github.com/flanksource/mission-control/blob/main/cmd/catalog.go#L27

`component_config_traverse=72143d48-da4a-477f-bac11e9decf188a6
 direction=outgoing type=Kubernetes::Pod,Kubernetes::Node`

types/resource_selector.go
```diff
@@ -9,13 +9,26 @@ import (
...
 	"gorm.io/gorm/clause"
 	"gorm.io/gorm/schema"
 	"k8s.io/apimachinery/pkg/fields"
 	"k8s.io/apimachinery/pkg/labels"
 )
 
+type ComponentConfigTraversalArgs struct {
+	ComponentID string         `yaml:"component_id,omitempty" json:"component_id,omitempty"`
+	Direction   string         `yaml:"direction,omitempty" json:"direction,omitempty"`
+	Types       pq.StringArray `yaml:"types,omitempty" json:"types,omitempty"`
```
moshloop, line 23: should use resourceSelector.types instead

query/resource_selector.go
yashmehrotra: okay, updating

types/resource_selector.go
yashmehrotra: These types go directly into the postgres function, I'll see if the `Items` type work else I'll derive that from pq.StringArray

query/resource_selector.go
```diff
@@ -224,24 +224,71 @@ func SetResourceSelectorClause(ctx context.Context, resourceSelector types.Resou
...
 
 	return query, nil
 }
 
+func setSearchQueryParams(rs *types.ResourceSelector) {
+	if rs.Search == "" {
+		return
+	}
+
+	queries := strings.Split(rs.Search, ";")
```
moshloop, line 269: Why not a space as the separator ?

query/resource_selector.go
```diff
@@ -224,24 +224,71 @@ func SetResourceSelectorClause(ctx context.Context, resourceSelector types.Resou
...
+	for _, q := range queries {
+		items := strings.Split(q, "=")
+		if len(items) != 2 {
+			continue
+		}
+
+		switch items[0] {
+		case "component_config_traverse":
```
moshloop, line 277: Can we add in the other fields that are currently in the cli catalog query ?

yashmehrotra: Added

query/resource_selector.go
yashmehrotra: Changed to space

query/resource_selector.go
```diff
@@ -224,24 +247,62 @@ func SetResourceSelectorClause(ctx context.Context, resourceSelector types.Resou
...
+			// search: component_config_traverse=72143d48-da4a-477f-bac1-1e9decf188a6,outgoing,Kubernetes::Pod,Kubernetes::Node
+			// Args should be componentID, direction and types (compID,direction,type1,type2,type3)
+			args := strings.SplitN(items[1], ",", 3)
+			if len(args) == 3 {
+				rs.Functions.ComponentConfigTraversal = &types.ComponentConfigTraversalArgs{
+					ComponentID: args[0],
+					Direction:   args[1],
+					Types:       types.Items(strings.Split(args[2], ",")),
```
moshloop, line 281: why not just use the main selector types for this, 

i.e. 
```
component_config_traverse=72143d48-da4a-477f-bac1-1e9decf188a6. 
 direction=outgoing type=Kubernetes::Pod,Kubernetes::Node
```

yashmehrotra: Yeah, makes sense

PR #1095, opened 2024-09-25
chore: add deleted_at column to permissions

mainawycliffe:


moshloop:
I don't think we would soft-delete permissions, they should just be deleted

PR #1107, opened 2024-09-29
chore: update component config traversal lookup

yashmehrotra:


views/005_component_views.sql
```diff
@@ -147,23 +147,18 @@ CREATE OR REPLACE VIEW component_types AS
...
-                5
-            )
+        WITH config_id_paths AS (
+            SELECT config_items.id
+            FROM config_items
+            WHERE path LIKE '%' || (SELECT config_id FROM components WHERE components.id = $1::UUID) || '%'
         )
         SELECT components.id 
         FROM components 
-        WHERE config_id IN (SELECT related_config_ids.id FROM related_config_ids);
```
moshloop, line 167: I think a join will be faster: https://dba.stackexchange.com/questions/91247/optimizing-a-postgres-query-with-a-large-in

yashmehrotra: Done 👍🏽

Issue #1114, opened 2024-10-01
In configs view, join with materialized config_change_analysis count 


yashmehrotra:
- Also change MV to only show with status = 'open'

PR #1115, opened 2024-10-01
chore: use config_change_analysis_count in configs view

yashmehrotra:
Fixes: https://github.com/flanksource/duty/issues/1114


views/006_config_views.sql
```diff
@@ -3,6 +3,66 @@ DROP VIEW IF EXISTS configs CASCADE;
...
+WITH type_counts AS (
+    SELECT 
+        ca.config_id, 
+        ca.analysis_type, 
+        COUNT(*) AS type_count
+    FROM 
+        config_analysis ca
+    WHERE 
+        ca.status = 'open' AND
+        ca.last_observed >= NOW() - INTERVAL '7 days'
```
moshloop, line 17: I think we can remove the last_observed and then the index would be on config_id with a clause of status = 'open'

yashmehrotra: So always aggregate all the analysis that are open ?

PR #1116, opened 2024-10-02
chore: remove analysis since from summary request and add 3d materialized view

yashmehrotra:
Fixes: https://github.com/flanksource/mission-control/issues/1479

### DIFF
--- a/job/catalog.go
+++ b/job/catalog.go
@@ -4,6 +4,10 @@ import (
 	"github.com/flanksource/duty/context"
 )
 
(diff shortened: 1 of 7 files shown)


job/catalog.go
```diff
@@ -4,6 +4,10 @@ import (
 	"github.com/flanksource/duty/context"
 )
 
+func RefreshConfigItemAnalysisChangeCount3d(ctx context.Context) error {
```
moshloop, line 7: Can we use `config_item_summary_7d` instead ?

PR #1123, opened 2024-10-09
fix: delete_old_config_items

adityathebe:
resolves: https://github.com/flanksource/mission-control/issues/1501


moshloop:
I don't think we should be ignoring them, If there is a silence on a resource it shouldn't prevent it fromt being deleted - so I think we just need to update the cascase delete option

PR #1125, opened 2024-10-11
feat: allow namespace/name for scope in resource selector

yashmehrotra:
For using catalog lookup in k8s topology chart, we need to be able to distinguish b/w clusters, currently we do that via kubeconfig using kubernetes lookup.

We need to allow scope to be selectable by namespace/name to have the same functionality

### DIFF
--- a/query/resource_selector.go
+++ b/query/resource_selector.go
@@ -178,8 +178,17 @@ func SetResourceSelectorClause(ctx context.Context, resourceSelector types.Resou
 		query = query.Where("status IN ?", resourceSelector.Statuses)
 	}
 
+	agentID, err := getAgentID(ctx, resourceSelector.Agent)
+	if err != nil {
+		return nil, err
+	}
+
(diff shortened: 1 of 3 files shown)


query/resource_selector.go
```diff
@@ -480,3 +484,42 @@ func propertySelectorRequirementToSQLClause(q *gorm.DB, r labels.Requirement) *g
...
+	case "config_items":
+		scopeTable = "config_scrapers"
+	case "components":
+		scopeTable = "topologies"
+	default:
+		return "", api.Errorf(api.EINVALID, "scope is not supported for %s", table)
+	}
+
+	var id string
+	if err := ctx.DB().Table(scopeTable).Select("id").Where("name = ? AND namespace = ?", name, namespace).Find(&id).Error; err != nil {
```
moshloop, line 519: Shouldn't this be by agent ?

PR #1128, opened 2024-10-15
feat: use client-go leader election

adityathebe:


leader/election.go
```diff
@@ -1,139 +1,131 @@
...
+				updateLeaderLabel(ctx, true)
+				onLead(leadCtx)
 			},
-			Spec: v1.LeaseSpec{
-				HolderIdentity:       &identity,
-				LeaseDurationSeconds: lo.ToPtr(int32(leaseDuration.Seconds())),
-				AcquireTime:          &now,
-				RenewTime:            &now,
+			OnStoppedLeading: func() {
+				updateLeaderLabel(ctx, false)
```
moshloop, line 99: I don't think we can rely on this to add a label - if the pod is killed or partitions it will never run - The leader should first add its own label, and then remove the label from other others

moshloop: I think we should also stop and start all schedules here

leader/election.go
```diff
@@ -1,139 +1,204 @@
...
-				AcquireTime:          &now,
-				RenewTime:            &now,
-			},
-		}
-		_, err = leases.Create(ctx, lease, metav1.CreateOptions{})
-		if err != nil {
-			return false, "", err
+
+	// Get the ReplicaSet owner reference
+	var replicaSetName string
```
moshloop, line 182: This would not work for rollouts, the pods would be in different replicasets = we just need to list pods with the leader label.

leader/election.go
```diff
@@ -1,139 +1,204 @@
...
-				// sleep for just under half the lease duration before trying to renew
-				time.Sleep(leaseDuration/2 - time.Second*10)
-			}
-		}()
+	if n, err := getPodNamespace(); err == nil {
+		podNamespace = n
 	}
 
+	// Not sure if this is a very reliable way to get the service name
+	service = strings.Split(hostname, "-")[0]
```
moshloop, line 69: I think pass this through to register

leader/election.go
```diff
@@ -1,139 +1,204 @@
...
+		pods, err := getAllReplicas(ctx, hostname)
+		if err != nil {
+			return retry.RetryableError(fmt.Errorf("failed to get replicas: %w", err))
+		}
+
+		for _, pod := range pods.Items {
+			var payload string
+			if pod.Name == hostname {
+				ctx.Infof("adding leader metadata from pod: %s", pod.Name)
+				payload = `{"metadata":{"labels":{"leader":"true"}}}`
```
moshloop, line 151: i think include the service name in the label e.g. `canary-checker/leader: true`

leader/election.go
```diff
@@ -1,139 +1,179 @@
...
-	renewTime := lease.Spec.RenewTime.Time
-	if time.Since(renewTime) > leaseDuration {
-		ctx.Infof("Lease %s held by %s expired", leaseName, *lease.Spec.HolderIdentity)
-		if err := leases.Delete(ctx, leaseName, metav1.DeleteOptions{}); err != nil {
-			ctx.Infof("failed to delete leases %s: %v", leaseName, err)
+
+		pods := lo.Map(podList.Items, func(p corev1.Pod, _ int) string { return p.Name })
+		pods = append(pods, hostname)
+
+		fmt.Println("Pods ", len(pods))
```
moshloop:
```suggestion
	
```

leader/election.go
```diff
@@ -1,139 +1,179 @@
...
-	if attempt > 0 {
-		time.Sleep(time.Duration(rand.Intn(5000)) * time.Millisecond)
-	}
-	if attempt >= ctx.Properties().Int("leader.lease.attempts", 3) {
-		return false, "", fmt.Errorf("failed to acquire lease %s after %d attempts", leaseName, attempt)
+func Register(
+	ctx context.Context,
+	app string,
+	namespace string,
+	service string,
```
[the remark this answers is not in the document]
adityathebe: app would always be "config-db" which is hardcoded in `flanksource/config-db` whereas service was meant to be the actual kubernetes service name.

Although, the we could just use the app name as the lease name.

leader/election.go
```diff
@@ -1,139 +1,200 @@
...
+				updateLeaderLabel(ctx, app)
+				onLead(leadCtx)
 			},
-			Spec: v1.LeaseSpec{
-				HolderIdentity:       &identity,
-				LeaseDurationSeconds: lo.ToPtr(int32(leaseDuration.Seconds())),
-				AcquireTime:          &now,
-				RenewTime:            &now,
+			OnStoppedLeading: func() {
+				for _, k := range echo.Crons.Items() {
```
moshloop, line 107: don't we also need to start the crons again when we become leader ?

adityathebe: starting the jobs had a few pre-processing before the cron was started.
So the operator in config-db would start those jobs in the `onLead` callback.

Although upon closer inspection, I see that's a one time thing.

PR #1139, opened 2024-10-21
fix: don't run cron on schedule

adityathebe:
Need to explicitly start the cron to support multiple replicase (leader election).

This way multiple replicas can register the job without starting it and let the leader election start/stop the cron schedulers.


moshloop:
@adityathebe isn't this a breaking change? What happens if leader election is not supported/enabled ?

adityathebe:
@moshloop if not in election-leader mode, we'll have to call `cron.Start()` from canary-checker, config-db ,...

moshloop:
@adityathebe can we not make this backwards compatible - i.e.  if leaderElectionEnabled {jobs.StartOnSchedule = false}

PR #1162, opened 2024-10-28
feat: RLS 

adityathebe:


models/permission_test.go
```diff
@@ -33,6 +34,13 @@ func TestPermission_Condition(t *testing.T) {
 			perm:     Permission{},
 			expected: "",
 		},
+		{
+			name: "agents",
```
moshloop, line 38: Don't we need one for tags

tests/rls_test.go
```diff
@@ -0,0 +1,78 @@
...
+				}),
+				Entry("correct agent", testCase{
+					name:          "correct agent",
+					jwtClaims:     `{"tags": {"cluster": "testing-cluster"}, "agents": ["00000000-0000-0000-0000-000000000000"]}`,
+					expectedCount: &totalConfigs,
+				}),
+				Entry("correct tag", testCase{
+					name:          "correct tag",
+					jwtClaims:     `{"tags": {"account": "flanksource"}, "agents": ["10000000-0000-0000-0000-000000000000"]}`,
+					expectedCount: &numConfigsWithFlanksourceTag,
```
moshloop, line 73: Are these not going to be 0 as no test data is inserted ?

adityathebe: the testdata are the dummy configs we have inn suite setup

moshloop:
@adityathebe Can we benchmark this @ 10k, , 25k, 50k, and 100k and add the results with and without RLS to a BENCHMARK.md ?

PR #1176, opened 2024-11-02
chore: add error ids on fkey error and process configs before components

yashmehrotra:


upstream/jobs.go
```diff
@@ -263,9 +269,12 @@ func reconcileTable(ctx context.Context, config UpstreamConfig, table pushableTa
 			})
 			failedItems := lo.Filter(items, func(item models.DBTable, _ int) bool {
 				_, ok := failedOnes[item.PK()]
+				if ok {
+					fkFailed.IDs = append(fkFailed.IDs, item.PK())
```
moshloop, line 273: Can we limit these to first 10 only

yashmehrotra: Added that logic in its json Marshal function

PR #1194, opened 2024-11-11
feat: notification wait for in send history

adityathebe:


models/notifications.go
```diff
@@ -68,13 +69,19 @@ type NotificationSendHistory struct {
 	CreatedAt      time.Time `json:"created_at" time_format:"postgres_timestamp"`
 	Status         string    `json:"status,omitempty"`
 
+	// payload holds in original event properties for delayed/pending notifications
+	Payload types.JSONStringMap `json:"payload,omitempty"`
+
+	// The duration to delay sending after its creation.
+	Delay *time.Duration `json:"delay,omitempty"`
```
moshloop, line 76: I think this should be a timeout timestamp so that date math is not required on query - re the payload - I don't think we should be saving it, it should be re-hydrated from the config id / change it / check_id so that the notification sends out up to date info

adityathebe: we do need to save the payload as it contains
- resource id (config/component/check)
- original event data to get the correct cel env.
- recipient details

we will still do send out up-to date info.

PR #1201, opened 2024-11-21
fix: group by for catalog summary

yashmehrotra:
Fixes: https://github.com/flanksource/duty/issues/1170


moshloop:
@yashmehrotra can we add a test for this ?

PR #1224, opened 2024-12-19
feat: config query grammar

moshloop:


query/grammar/grammar_test.go
```diff
@@ -0,0 +1,17 @@
...
+import (
+	"github.com/flanksource/commons/logger"
+	. "github.com/onsi/ginkgo/v2"
+	. "github.com/onsi/gomega"
+)
+
+var _ = Describe("grammer", func() {
+
+	It("parses", func() {
+		result, err := ParsePEG("john:doe metadata.name=bob metadata.name!=harry spec.status.reason!=\"failed reson\"   -jane johnny type!=pod type!=replicaset  namespace!=\"a,b,c\"")
```
moshloop, line 12: Can we expand these tests out ?

tests/query_resource_selector_test.go
```diff
@@ -282,3 +286,102 @@ var _ = ginkgo.Describe("Resoure Selector limits", ginkgo.Ordered, func() {
...
+	// = , != , item in list, item not in list, prefix, suffix, date operations (created_at, updated_at), agent query
+	testData := []struct {
+		description string
+		query       string
+		expectedIDs []uuid.UUID
+		resource    string
+	}{
+		{
+			description: "config item query",
+			query:       `name="node-b" type="Kubernetes::Node"`,
```
moshloop, line 304:
```suggestion
			query:       `name=node-b type=Kubernetes::Node`,
```

moshloop: can we also add a `name=node-b type=kubernetes::node` (case insensitivity)

yashmehrotra: Doing the suggested will cause the test to fail as things get parsed incorrectly

PR #1236, opened 2024-12-28
feat: table schemas for permission & permission group

adityathebe:
related: https://github.com/flanksource/mission-control/issues/1700

### DIFF
--- a/types/resource_selector.go
+++ b/types/resource_selector.go
@@ -193,14 +193,23 @@ func (q QueryField) ToClauses() ([]clause.Expression, error) {
 	return clauses, nil
 }
 
-func (c ResourceSelector) IsEmpty() bool {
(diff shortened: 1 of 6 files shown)


types/resource_selector.go
```diff
@@ -84,6 +84,15 @@ func (c ResourceSelector) IsEmpty() bool {
 		len(c.FieldSelector) == 0
 }
 
+// A wildcard resource selector is one where it just has the name field set to '*'
+func (c ResourceSelector) Wildcard() bool {
+	return c.allEmptyButName() && c.Name == "*"
+}
```
adityathebe, line 90: @moshloop made this a global thing

views/036_permission.sql
```diff
@@ -0,0 +1,48 @@
+UPDATE
+  permissions
+SET
+  subject = COALESCE(team_id, person_id, notification_id)
```
moshloop, line 4: Don't we need a subject type here ?

PR #1249, opened 2025-01-06
feat: playbooks as resource selectable

adityathebe:


models/playbooks.go
```diff
@@ -17,6 +17,8 @@ import (
 	"k8s.io/apimachinery/pkg/labels"
 )
 
+var AllowedColumnFieldsInPlaybooks = []string{"category"}
```
moshloop, line 20: Can we not add a PlaypookQueryModel in here: https://github.com/flanksource/duty/blob/main/query/models.go

adityathebe: I've added PlaybookQueryModel. 

But we still need `AllowedColumnFieldsInPlaybooks`. The query model seems to be used only for the `.search` field.

https://github.com/flanksource/duty/blob/main/models/config.go#L71

PR #1268, opened 2025-01-15
fix: resource selector cache scoped per user when RLS is enabled

adityathebe:
related: https://github.com/flanksource/mission-control/issues/1750


query/resource_selector.go
```diff
@@ -376,6 +376,12 @@ func queryResourceSelector(ctx context.Context, limit int, resourceSelector type
 	}
 
 	hash := fmt.Sprintf("%s-%s-%d", table, resourceSelector.Hash(), limit)
+
+	// NOTE: When RLS is enabled, we need to scope the cache per user.
```
moshloop, line 380: Shouldn't this be scoped to the rls tags/agents ?

Issue #1300, opened 2025-01-28
Resource selector search properties


adityathebe:
We have removed properties search from the field selector.

`properties.region = us-east-1` doesn't currently work because properties, unlike other jsonb columns, is an array.

```json
[
  {
    "name": "region",
    "text": "us-east-1"
  }
]

```

adityathebe:
https://github.com/flanksource/duty/pull/1308/files

PR #1308, opened 2025-01-30
feat: unkeyed search on tags & properties

adityathebe:
resolves: https://github.com/flanksource/duty/issues/1305

```sql
incident_commander> EXPLAIN (ANALYZE, BUFFERS) SELECT name, type, tags
 FROM config_items
 WHERE  tags_values ?| ARRAY['immich']
+----------------------------------------------------------------------------------------------------------------------------------------+
| QUERY PLAN                                                                                                                             |
|----------------------------------------------------------------------------------------------------------------------------------------|
| Bitmap Heap Scan on config_items  (cost=12.95..122.39 rows=31 width=101) (actual time=0.014..0.041 rows=31 loops=1)                    |
|   Recheck Cond: (tags_values ?| '{immich}'::text[])                                                                                    |
|   Heap Blocks: exact=28                                                                                                                |
|   Buffers: shared hit=32                                                                                                               |
|   ->  Bitmap Index Scan on idx_config_items_tags_values  (cost=0.00..12.94 rows=31 width=0) (actual time=0.010..0.010 rows=31 loops=1) |
|         Index Cond: (tags_values ?| '{immich}'::text[])                                                                                |
|         Buffers: shared hit=4                                                                                                          |
| Planning:                                                                                                                              |
|   Buffers: shared hit=134                                                                                                              |
| Planning Time: 0.169 ms                                                                                                                |
| Execution Time: 0.048 ms                                                                                                               |
+----------------------------------------------------------------------------------------------------------------------------------------+
```

vs

```sql
incident_commander> EXPLAIN (ANALYZE, BUFFERS) SELECT name, type, tags
 FROM config_items
 WHERE EXISTS (SELECT 1 FROM jsonb_each_text(tags) WHERE value = 'immich')
+-------------------------------------------------------------------------------------------------------------------------+
| QUERY PLAN                                                                                                              |
|-------------------------------------------------------------------------------------------------------------------------|
| Seq Scan on config_items  (cost=0.00..10967.96 rows=3868 width=101) (actual time=0.659..17.062 rows=31 loops=1)         |
|   Filter: (SubPlan 1)                                                                                                   |
|   Rows Removed by Filter: 7700                                                                                          |
|   Buffers: shared hit=1200 dirtied=16                                                                                   |
|   SubPlan 1                                                                                                             |
|     ->  Function Scan on jsonb_each_text  (cost=0.00..1.25 rows=1 width=0) (actual time=0.002..0.002 rows=0 loops=7731) |
[...]

PR #1309, opened 2025-01-30
Feat/properties lookup

adityathebe:
resolves: https://github.com/flanksource/duty/issues/1300


adityathebe:
Implemented here: https://github.com/flanksource/duty/pull/1308/files

PR #1314, opened 2025-02-03
fix: latency calculation in check

yashmehrotra:
Fixes: https://github.com/flanksource/canary-checker/issues/1986

### DIFF
--- a/query/check_details.go
+++ b/query/check_details.go
@@ -3,14 +3,15 @@ package query
 import (
 	"fmt"
 	"net/url"
+	"slices"
 	"strings"
 	"time"
 
-	"github.com/asecurityteam/rolling"
 	"github.com/flanksource/commons/duration"
 	"github.com/flanksource/duty/context"
 	"github.com/flanksource/duty/types"
 	"github.com/samber/lo"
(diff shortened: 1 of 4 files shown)


query/check_details.go
```diff
@@ -189,15 +190,17 @@ ORDER BY time
 		}
 		uptime.Failed += datapoint.Failed
 		uptime.Passed += datapoint.Passed
-		latencies.Append(float64(datapoint.Duration))
+		latencies = append(latencies, float64(datapoint.Duration))
```
moshloop, line 193: Latencies will grow for as long canary-checker ius running:

This is what claude produced for a circular buffer, that segements latencies into buckets and then drops buckets after they expire: https://gist.github.com/moshloop/45308b7f3b48bd1a1a9b7ef0a54e4056 

Although I suspect this will cause a significant amount of CPU in environments with hundreds to thousands of canaries.

The reason we originally chose the rolling implementation is that it is very fast, and approximately right.

yashmehrotra: This is just for the graph api call, not the CRD status latencies. For this, we query check_statuses duration and then take the percentlie latency from that set

PR #1316, opened 2025-02-03
feat: configurable duration for slow queries

adityathebe:


gorm/logger.go
```diff
@@ -51,8 +51,7 @@ type Logger interface {
 }
 
 type Config struct {
-	// SlowThreshold in nanoseconds
-	SlowThreshold             int64
```
moshloop, line 55: This is for all DB query

I am thinking of:

```
query.log.slowDuration=100ms
query.log=debug
```

where INFO:
`[SLOW 100ms:100]config grammar`
DEBUG
`[100ms:100]config grammar`
TRACE
`[100ms:100]config grammar -> generated sql`

PR #1330, opened 2025-02-15
feat: kubernetes client cache

yashmehrotra:


connection/kubernetes.go
```diff
@@ -51,6 +51,33 @@ type KubernetesConnection struct {
...
+		return fmt.Sprintf("gke=%s/%v", c.GKE.Cluster, c.GKE.ToModel())
+	}
+	if c.CNRM != nil {
+		return fmt.Sprintf("cnrm=%s/%v", c.CNRM.ClusterResource, c.GKE.ToModel())
+	}
+	return "local"
+}
+
+func (c KubernetesConnection) CanExpire() bool {
+	return c.EKS != nil || c.GKE != nil || c.CNRM != nil
```
moshloop, line 80: It can also expire if reading kubeconfig from a secret

context/kuberetes_client.go
```diff
@@ -0,0 +1,43 @@
...
+	}
+	_, rc, err := c.Connection.Populate(ctx, true)
+	if err != nil {
+		return fmt.Errorf("error refreshing kubernetes client: %w", err)
+	}
+
+	// Update rest config in place for easy reuse
+	c.Config.Host = rc.Host
+	c.Config.TLSClientConfig = rc.TLSClientConfig
+	c.Config.BearerToken = rc.BearerToken
```
moshloop, line 32: We should probably try and decode this as a JWT token and use the expiry date defined. 

If reading a kubeconfig from a secret then it could be usung BearerTokenFile / Username / Password etc so we need to handle that case.

context/kuberetes_client.go
```diff
@@ -0,0 +1,43 @@
...
+	*dutyKubernetes.Client
+	Connection KubernetesConnection
+	expiry     time.Time
+}
+
+func (c *KubernetesClient) SetExpiry(d time.Duration) {
+	c.expiry = time.Now().Add(d)
+}
+
+func (c *KubernetesClient) RefreshWithExpiry(ctx Context, d time.Duration) error {
```
moshloop, line 20: Expiry is something that should come from connection populate, or the JWT token expirty it should. be user specified.

context/context.go
```diff
@@ -350,62 +356,56 @@ func (k Context) Pool() *pgxpool.Pool {
...
 func (k *Context) KubeAuthFingerprint() string {
-	rs := k.KubernetesRestConfig()
-	if rs == nil {
+	kc, _ := k.Kubernetes()
+	if kc == nil {
 		return ""
 	}
-
-	return hash.Sha256Hex(fmt.Sprintf("%s/%s/%s/%s/%s/%s",
-		rs.Host,
```
moshloop, line 359: Might be missing APIserver path etc.. 

This hash doesn't uniquely identify the server, if connections with the same server, but a different bearer token should have the same hash and use the same cache.

connection/kubernetes.go
yashmehrotra: Updated

context/kuberetes_client.go
yashmehrotra: > If reading a kubeconfig from a secret then it could be usung BearerTokenFile / Username / Password etc so we need to handle that case.
Updated

> We should probably try and decode this as a JWT token and use the expiry date defined.
So just try to decode and see if it has an expiry ? It won't always be a jwt right?

context/context.go
```diff
@@ -350,62 +356,64 @@ func (k Context) Pool() *pgxpool.Pool {
...
+	_ = k8sclientcache.Set(k, connHash, client)
+	return client.Client, nil
 }
 
-func (k *Context) WithKubeconfig(input types.EnvVar) (*Context, error) {
-	if k.GetNamespace() == "" {
-		return nil, k.Oops().Errorf("namespace is required")
-	}
+func (k Context) WithLocalKubernetes(client *dutyKubernetes.Client) Context {
+	return k.WithValue("localKubernetes", client)
```
moshloop, line 400: Shouldn't this be static ? or include a getOrCreate on the cache ?

PR #1332, opened 2025-02-18
chore: explicitly give data path permissions and use pg15

yashmehrotra:


start.go
```diff
@@ -202,14 +202,19 @@ func embeddedDB(database, connectionString string, port uint32) (string, func()
...
 	logger.Infof("Starting embedded postgres server at %s", embeddedPath)
 
 	embeddedPGServer := embeddedpostgres.NewDatabase(embeddedpostgres.DefaultConfig().
 		Port(port).
-		DataPath(path.Join(embeddedPath, "data")).
+		DataPath(dataPath).
 		RuntimePath(path.Join(embeddedPath, "runtime")).
 		BinariesPath(path.Join(embeddedPath, "bin")).
-		Version(embeddedpostgres.V14).
+		Version(embeddedpostgres.V15).
```
adityathebe, line 217: I think now we're targetting v16.

moshloop: I think we also need todo a version checkon on any existing db as it will not auto upgrade

yashmehrotra: Updated flow

PR #1339, opened 2025-02-24
chore: remove custom clientset from k8s

yashmehrotra:


yashmehrotra:
@moshloop Self merging this FYI

Issue #1238, opened 2024-12-31
Resource selector: type queries should be case insensitive


moshloop:

moshloop:
Resolve types by maintaining a distinct list of unique types, and then using matchItem from the user supplied value to list replacing it with the correct value(s). 
+ Make the prefix optional e.g. `podinfo type=pod`

PR #1344, opened 2025-02-26
feat: support case insensitive suffix type matching

adityathebe:
resolves: https://github.com/flanksource/duty/issues/1238

### DIFF
--- a/query/jobs.go
+++ b/query/jobs.go
@@ -29,9 +29,7 @@ var updateTypesCache = &job.Job{
 			return err
 		}
 
-		allTypesCache.lock.Lock()
-		allTypesCache.Types = types
-		allTypesCache.lock.Unlock()
+		allTypesCache.Swap(types)
 		return nil
 	},
 }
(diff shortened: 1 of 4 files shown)


query/models.go
```diff
@@ -18,6 +21,9 @@ import (
 	"gorm.io/gorm/clause"
 )
 
+// Maintained by a job
+var allTypesCache = cache.New(time.Hour*24, time.Hour*24)
```
yashmehrotra, line 25: Prefer using duty/cache

query/jobs.go
```diff
@@ -0,0 +1,40 @@
...
+	Name:          "UpdateTypesCache",
+	Schedule:      "@every 5m",
+	Retention:     job.RetentionFailed,
+	Singleton:     true,
+	JobHistory:    true,
+	RunNow:        true,
+	JitterDisable: true,
+	Fn: func(ctx job.JobRuntime) error {
+		var types []string
+		query := `SELECT type FROM config_items UNION SELECT type FROM components UNION SELECT type FROM checks`
```
moshloop, line 27: Shouldn't this be distinct ?

PR #1345, opened 2025-02-26
fix: notification fallback

adityathebe:


models/notifications.go
```diff
@@ -140,7 +141,8 @@ type NotificationSendHistory struct {
 	// Hash for grouping resources with same message
 	GroupByHash string `json:"group_by_hash,omitempty"`
 
-	IsFallback bool `json:"is_fallback,omitempty"`
+	// ID of the original send history that is this notification history is a fallback of.
+	ParentID *uuid.UUID `json:"parent_id,omitempty"`
```
moshloop, line 145: Shouldn't this also be used for grouping ?

adityathebe: I assume that's just a matter of adding the `parent_id` to the rest of the notification send histories in a group?

PR #1353, opened 2025-03-06
fix: kubernetes connection refresh

yashmehrotra:


context/kuberetes_client.go
```diff
@@ -50,22 +50,41 @@ func (c *KubernetesClient) SetExpiry(def time.Duration) {
...
 	c.Config.Username = rc.Username
 	c.Config.Password = rc.Password
 
+	logger.Infof("Host %s", rc.Host)
+	logger.Infof("BearerToken %+v", rc.BearerToken)
+	logger.Infof("Address of bt in Refresh is %p", &c.Client.Config.BearerToken)
+
+	if doRefresh {
+		logger.Infof("Refreshing client for host %s", rc.Host)
+		c.Client.Interface = client
```
moshloop, line 84: I think we can do this always as the client as already been created and is pretty cheap.

moshloop: Also https://github.com/flanksource/duty/blob/main/kubernetes/k8s.go#L55 that kube-client-cache is not expiry aware, so we could ask for a new client and it returns something that is 30m old that has already expired

yashmehrotra: That cache has 1 hour expiry, should I update it to never use the cache itself ?

yashmehrotra: Remove the kubeconfig based cache

PR #1355, opened 2025-03-10
feat: custom auth plugin for kubernetes

yashmehrotra:


pkg/kube/auth/auth.go
```diff
@@ -0,0 +1,480 @@
...
+	// interactive we want to make sure it's only called once.
+	mu          sync.Mutex
+	cachedCreds *credentials
+	exp         time.Time
+
+	// getCert makes Authenticator.cert comparable to support TLS config caching
+	getCert *transport.GetCertHolder
+	// dial is used for clients which do not specify a custom dialer
+	// it is comparable to support TLS config caching
+	dial *transport.DialHolder
```
moshloop, line 220:
```suggestion
	// Set by the config
	group              schema.GroupVersion
	cluster            *clientauthentication.Cluster
	provideClusterInfo bool

	// Used to avoid log spew by rate limiting install hint printing. We didn't do
	// this by interval based rate limiting alone since that way may have prevented
	// the install hint from showing up for kubectl users.
	sometimes   *sometimes
	installHint string


	// connTracker tracks all connections opened that we need to close when rotating a client certificate
	connTracker *connrotation.ConnectionTracker

	// Cached results.
	//
	// The mutex also guards calling the plugin. Since the plugin could be
	// interactive we want to make sure it's only called once.
	mu          sync.Mutex
	cachedCreds *credentials
	exp         time.Time

	// getCert makes Authenticator.cert comparable to support TLS config caching
	getCert *transport.GetCertHolder
	// dial is used for clients which do not specify a custom dialer
	// it is comparable to support TLS config caching
	dial *transport.DialHolder
```

pkg/kube/auth/auth.go
```diff
@@ -0,0 +1,480 @@
...
+		return nil
+	}
+
+	return a.refreshCredsLocked()
+}
+
+// refreshCredsLocked executes the plugin and reads the credentials from
+// stdout. It must be called while holding the Authenticator's mutex.
+func (a *Authenticator) refreshCredsLocked() error {
+	interactive, err := a.interactiveFunc()
```
moshloop, line 353: This is where we need todo the callback rather than calling a binary

PR #1356, opened 2025-03-10
fix: setting up connections for non-existing configs or configs with no scrapeconfig

adityathebe:
Valid ids for configs & scrapers, before making the SQL call. else, it causes a SQL error and aborts any running transactions.

This caused, playbook to be in a crashloop.

resolves: https://github.com/flanksource/mission-control/issues/1919

### DIFF
--- a/connection/environment.go
+++ b/connection/environment.go
@@ -9,6 +9,8 @@ import (
 	"path/filepath"
 	"strings"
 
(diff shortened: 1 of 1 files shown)


moshloop:
@adityathebe should we not ensure that sql errors are handled and the playbook updated with with error status ?

PR #1358, opened 2025-03-11
feat: permissions_summary

adityathebe:
related: https://github.com/flanksource/flanksource-ui/issues/2440


adityathebe:
@moshloop I think we need to address this before this: https://github.com/flanksource/mission-control/issues/1926

PR #1401, opened 2025-04-10
feat: add resolution timestamp to notification group resources

adityathebe:


migrate/migrate.go
```diff
@@ -44,6 +44,10 @@ func RunMigrations(pool *sql.DB, config api.Config) error {
 		config.SkipMigrationFiles = append(config.SkipMigrationFiles, "034_rls_enable.sql", "035_rls_disable.sql")
 	}
 
+	// Always run this to prevent queries from failing after an upgrade.
+	// It's idempotent, so it's safe to run multiple times.
+	config.MustRun = append(config.MustRun, "037_notification_group_resources.sql")
```
moshloop, line 49: Can we do this with a `-- run: always` tag similar to how we do dependsOn

Issue #1415, opened 2025-04-22
Kubernetes client `QueryResources` doesn't fetch cluster scoped resources (like namespace)


adityathebe:
```go
func (c *Client) QueryResources(ctx context.Context, selector types.ResourceSelector) ([]unstructured.Unstructured, error) {
```

PR #1416, opened 2025-04-22
fix: kubenetes QueryResource support for non-namespaced resources

adityathebe:
resolves: https://github.com/flanksource/duty/issues/1415

### DIFF
--- a/kubernetes/dynamic.go
+++ b/kubernetes/dynamic.go
@@ -474,6 +474,20 @@ func (c *Client) QueryResources(ctx context.Context, selector types.ResourceSele
 
 	var resources []unstructured.Unstructured
 	for _, kind := range selector.Types {
+		if strings.ToLower(kind) == "namespace" && selector.IsMetadataOnly() {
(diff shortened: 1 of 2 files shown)


kubernetes/dynamic.go
```diff
@@ -468,58 +474,47 @@ func (c *Client) QueryResources(ctx context.Context, selector types.ResourceSele
...
-				}
-				namespaces = []string{name}
-			} else {
-				namespaces = []string{""}
-			}
+		isClusterScoped := rm.Scope.Name() == meta.RESTScopeNameRoot
+
+		var namespaces []string
+		if isClusterScoped {
+			namespaces = []string{""}
```
moshloop, line 486: We seem to have dropped the selector IsMetadataOonly support which means we hit the API to get the namespace but never actually use anything besides for the name

adityathebe: I've added it back but I couldn't see it being used anywhere

PR #1431, opened 2025-05-12
feat: pubsub types

yashmehrotra:
Fixes: #1430


pubsub/types.go
```diff
@@ -0,0 +1,123 @@
...
+	Subscription string `json:"subscription"`
+}
+
+func (p PubSubConfig) String() string {
+	return fmt.Sprintf("gcppubsub://projects/%s/subscriptions/%s", p.ProjectID, p.Subscription)
+}
+
+// +kubebuilder:object:generate=true
+type NATSConfig struct {
+	URL     string `json:"url,omitempty"`
```
yashmehrotra, line 78: @moshloop Should this be string or the URL type we use in RabbitMQ since the CRD seemed to accept a URL object

moshloop (approved):

PR #1437, opened 2025-05-16
feat: new access logs table

adityathebe:


models/config_access.go
```diff
@@ -134,19 +128,31 @@ func (e ConfigAccess) PK() string {
...
 }
 
 func (e UserConfigAccessSummary) TableName() string {
 	return "user_config_access_summary"
 }
+
+type ConfigAccessLog struct {
+	ConfigID       uuid.UUID `json:"config_id" gorm:"primaryKey"`
+	ExternalUserID uuid.UUID `json:"external_user_id" gorm:"primaryKey"`
+	ScraperID      uuid.UUID `json:"scraper_id" gorm:"primaryKey"`
```
moshloop, line 152: Why do we need scraper Id ?

We should also add `MFA bool` and `properties jsonb`

adityathebe: > Why do we need scraper Id ?

tying up all config access tables to a scraper to help during cleanup/deletion.

PR #1467, opened 2025-06-26
feat(search): add glob support and fix match query tests

yashmehrotra:
Depends on https://github.com/flanksource/commons/pull/267

### DIFF
--- a/query/grammar/filters.go
+++ b/query/grammar/filters.go
@@ -17,7 +17,7 @@ const (
 )
 
 type expressions struct {
-	In     []interface{}
+	In     []any
 	Prefix []string
 	Suffix []string
 	Glob   []string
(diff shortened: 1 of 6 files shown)


query/grammar/filters.go
```diff
@@ -111,8 +127,7 @@ func ParseFilteringQueryV2(query string, decodeURL bool) (FilteringQuery, error)
 		return result, nil
 	}
 
-	items := strings.Split(query, ",")
-	for _, item := range items {
+	for item := range strings.SplitSeq(query, ",") {
```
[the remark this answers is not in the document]
yashmehrotra: This is wrong, the go version to use 1.24, they allow it now due to it being an iter.Seq

Can you recheck

tests/query_resource_selector_test.go
```diff
@@ -765,6 +785,10 @@ var _ = ginkgo.Describe("Resoure Selector with PEG", ginkgo.Ordered, func() {
 	ginkgo.Describe("peg search", func() {
 		for _, tt := range testData {
 
+			if !tt.focus {
```
moshloop, line 788: doesn't this skip the tests

yashmehrotra: removed

PR #1472, opened 2025-07-01
feat: views reconciliation

adityathebe:
related: https://github.com/flanksource/mission-control/issues/2163

* remove foreign key view_panels to view table as we only push view_panels to the upstream without pushing the view. The view on upstream will have its own id


tests/upstream_test.go
```diff
@@ -59,6 +65,7 @@ var _ = ginkgo.Describe("Reconcile Test", ginkgo.Ordered, ginkgo.Label("slow"),
 
 		e.Use(upstream.AgentAuthMiddleware(cache.New(time.Hour, time.Hour)))
 		e.POST("/upstream/push", upstream.NewPushHandler(nil))
+		e.POST("/upstream/check-view", upstream.CheckViewHandler)
```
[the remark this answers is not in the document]
adityathebe: We aren't querying `information_schema.

We just match view.spec.columnDefinition in the agent and in the upstream.

models/common.go
```diff
@@ -148,3 +150,87 @@ type TaggableModel interface {
...
+			}
+
+		default:
+			// do nothing
+		}
+	}
+
+	// Log collected warnings
+	for warning := range warnings {
+		logger.Warnf("postProcessViewRows: %s", warning)
```
moshloop, line 232: These should be returned with the query

tests/testdata/views/deployments.yaml
```diff
@@ -0,0 +1,45 @@
+apiVersion: mission-control.flanksource.com/v1
+kind: View
+metadata:
+  name: deployments
+  namespace: mc
+  uid: 6d53ff5a-13ec-4049-8a97-adad0fcd97ca
```
moshloop, line 6:
```suggestion
```

tests/testdata/views/deployments.yaml
```diff
@@ -0,0 +1,45 @@
...
+    configs:
+      - selector:
+          search: '@order=name'
+          types:
+            - Kubernetes::HelmRelease
+          tagSelector: namespace in (monitoring,media)
+        mapping:
+          application: row.name
+          namespace: row.tags.namespace
+          chart: row.config.status.history[0].chartName
```
moshloop, line 41: I think this would be nullsafe
```suggestion
          chart: jsonpath(row.config, '$.status.history[0].chartName')
```

tests/upstream_test.go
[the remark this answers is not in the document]
moshloop: The server side doesn't need to be cached, but we should cache client side as reconcole_all is every minute

upstream/controllers.go
```diff
@@ -168,3 +171,47 @@ func PingHandler(c echo.Context) error {
...
+	}
+
+	var viewModel models.View
+	if err := ctx.DB().Where("namespace = ? AND name = ?", req.Namespace, req.Name).Find(&viewModel).Error; err != nil {
+		return c.JSON(http.StatusNotFound, api.HTTPError{Err: err.Error(), Message: "view not found"})
+	} else if viewModel.ID == uuid.Nil {
+		return c.JSON(http.StatusNotFound, api.HTTPError{Err: "view not found", Message: "view not found"})
+	}
+
+	if !ctx.DB().Migrator().HasTable(viewModel.GeneratedTableName()) {
```
moshloop, line 196: Isn't this an information_schema query ?

tests/testdata/views/deployments.yaml
moshloop: `Select *,  configs.config->status.history[0].chartName`

upstream/client.go
```diff
@@ -120,6 +126,39 @@ func (t *UpstreamClient) push(ctx context.Context, method string, msg *PushData)
 	return nil
 }
 
+// ListViews returns all views from upstream with namespace,name pairs
+func (t *UpstreamClient) ListViews(ctx context.Context, views []ViewIdentifier) ([]ViewWithColumns, error) {
+	cacheKey := "columns-list"
```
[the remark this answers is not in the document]
adityathebe: nonsense

view/db.go
```diff
@@ -0,0 +1,142 @@
...
+	case ColumnTypeGauge:
+		return "JSONB"
+	default:
+		return "TEXT"
+	}
+}
+
+func ReadViewTable(ctx context.Context, table string) ([]Row, error) {
+	rows, err := ctx.DB().Raw(fmt.Sprintf("SELECT * FROM %s", table)).Rows()
+	if err != nil {
```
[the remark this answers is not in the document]
adityathebe: Fixed with 

```go
ctx.DB().Select("*").Table(table).Rows()
```

view/db.go
```diff
@@ -0,0 +1,142 @@
...
+	for _, col := range columns {
+		colDef := fmt.Sprintf("%s %s", col.Name, getPostgresType(col.Type))
+		columnDefs = append(columnDefs, colDef)
+	}
+
+	columnDefs = append(columnDefs, "agent_id UUID NOT NULL DEFAULT '00000000-0000-0000-0000-000000000000'::uuid")
+	columnDefs = append(columnDefs, "is_pushed BOOLEAN DEFAULT FALSE")
+
+	sql := fmt.Sprintf("CREATE TABLE IF NOT EXISTS %s (%s)", table, strings.Join(columnDefs, ", "))
+	return ctx.DB().Exec(sql).Error
```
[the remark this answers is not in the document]
adityathebe: In https://github.com/flanksource/mission-control/issues/2186, we'll move to atlas with schema defined in hcl

models/common.go
```diff
@@ -148,3 +151,119 @@ type TaggableModel interface {
...
+	ColumnTypeDuration ColumnType = "duration"
+	ColumnTypeInteger  ColumnType = "integer"
+	ColumnTypeJSONB    ColumnType = "jsonb"
+	ColumnTypeString   ColumnType = "string"
+)
+
+// ConvertViewRecordsToNativeTypes converts view cell to native go types
+func ConvertViewRecordsToNativeTypes(row map[string]any, columnDef map[string]ColumnType) (map[string]any, map[string][]string) {
+	// keep track of all the invalid types encountered per column
+	invalidTypesPerColumn := make(map[string][]string)
```
moshloop, line 170: We don't want an error per cell, only per row, if their is a casting issue it is likely to be an issue for all rows

view/db.go
```diff
@@ -0,0 +1,142 @@
...
+		}
+
+		viewRows = append(viewRows, viewRow)
+	}
+
+	return viewRows, nil
+}
+
+func InsertViewRows(ctx context.Context, table string, columns []ViewColumnDef, rows []Row) error {
+	if err := ctx.DB().Exec(fmt.Sprintf("TRUNCATE TABLE %s", table)).Error; err != nil {
```
moshloop, line 117: don't we need todo this in a transaction ?

adityathebe: > TRUNCATE isn't MVCC safe, so some clients in other transactions will see the table as empty.

I don't think that's true. From what I've observed, other clients just wait and see the newly committed rows.

We can use `DELETE FROM` which doesn't block.

view/db.go
```diff
@@ -0,0 +1,142 @@
...
+		}
+
+		viewRows = append(viewRows, viewRow)
+	}
+
+	return viewRows, nil
+}
+
+func InsertViewRows(ctx context.Context, table string, columns []ViewColumnDef, rows []Row) error {
+	if err := ctx.DB().Exec(fmt.Sprintf("DELETE FROM %s", table)).Error; err != nil {
```
moshloop, line 117: This is would cause a  large amount of churn in the WAL, especially if the rows are mostly static and/or only have 1 or 2 columns that get updated. 

I think we should upsert with a touched_at column, and then cleanup old rows at the end

PR #1479, opened 2025-07-08
chore: add job to reset is pushed

yashmehrotra:


upstream/jobs.go
```diff
@@ -352,3 +352,31 @@ func reconcileTable(ctx context.Context, config UpstreamConfig, table pushableTa
 		}
 	}
 }
+
+func ResetIsPushed(ctx context.Context) error {
+	overrides := map[string]string{
```
moshloop, line 357: can we make the interval configurble

moshloop: are we not missing config relationship ?

yashmehrotra: > can we make the interval configurble

Done

> are we not missing config relationship ?

No. we don't need to override it since it has created,updated and deleted fields

PR #1507, opened 2025-07-29
feat: add handling for agent config item

yashmehrotra:


views/002_seed.sql
```diff
@@ -12,6 +12,14 @@ BEGIN
     END IF;
 END $$;
 
+DO $$
+BEGIN
+    IF NOT EXISTS (SELECT FROM config_scrapers WHERE name = 'System') THEN
+        INSERT INTO config_scrapers (id, name, source, spec)
+        VALUES ('00000000-0000-0000-0000-000000000000', 'System', 'System', '{"schedule": "@every 30s", "system": true}');
```
moshloop, line 19:
```suggestion
        VALUES ('00000000-0000-0000-0000-000000000000', 'System', 'System', '{"schedule": "@every 5m", "system": true}');
```

PR #1534, opened 2025-08-19
feat: config locations

adityathebe:


dataquery/prometheus_test.go
adityathebe, line 1: Disabling these tests for now. prometheus.demo.prometheus.io is down

PR #1546, opened 2025-09-04
chore: convert catchall perm to perm per object

yashmehrotra:


rbac/enforcer.go
```diff
@@ -165,7 +165,17 @@ func PermsForUser(user string) ([]policy.Permission, error) {
 	}
 	var s []policy.Permission
 	for _, perm := range append(perms, implicit...) {
-		s = append(s, policy.NewPermission(perm))
+		p := policy.NewPermission(perm)
```
moshloop, line 168: What does the UI look like for this ?

yashmehrotra: Before
image


After
image

PR #1571, opened 2025-10-06
feat(RLS): support name

adityathebe:
we might not need this if we add Scope CRD


adityathebe:
we have access scope now

PR #1619, opened 2025-10-31
Refactor PostgreSQL LISTEN to use single connection

adityathebe:
Centralizes PostgreSQL LISTEN connection management by multiplexing notifications across multiple channels through a single connection with automatic reconnection on failure.

The Listen function handles the full connection lifecycle including retries and acquiring new connections. This allows callers throughout the codebase to simply call Listen without managing connections or handling failures themselves.

---

We initially wanted a ListenManager to allow applications to call Listen() from anywhere while sharing a single connection. The manager would handle multiplexing notifications from one connection to multiple consumers. However, this design has a fundamental timing problem with pgx's `conn.Conn().WaitForNotification`. Starting it during manager initialization causes all subsequent `Listen()` calls to fail with "conn busy" because `WaitForNotification` blocks the connection. Alternatively, starting it after the first `Listen()` call means every new registration requires interrupting `WaitForNotification` to execute the new LISTEN command, then restarting it - adding complexity and potential message loss.

The root issue is pgx's connection locking: WaitForNotification sets the connection to connStatusBusy and blocks indefinitely. Any other operation on that connection (including EXEC "LISTEN channel") immediately fails with "conn busy". This is by design - pgx connections cannot handle concurrent operations. 

The solution is to abandon the shared connection model entirely: each Listen()/ListenMany() call gets its own dedicated connection, and all LISTEN commands must be specified upfront before WaitForNotification starts.

Related: https://github.com/jackc/pgx/issues/195

Resolves: https://github.com/flanksource/duty/issues/1620

### DIFF
--- a/postq/pg/notify.go
+++ b/postq/pg/notify.go
@@ -32,7 +32,7 @@ func Listen(ctx context.Context, channel string, listener chan<- string) error {
 	backoff := retry.WithMaxDuration(DBReconnectMaxDuration, retry.NewExponential(DBReconnectBackoffBaseDuration))
 	return retry.Do(ctx, backoff, func(retryCtx gocontext.Context) error {
 		if err := listenLoop(ctx, map[string][]chan<- string{channel: {listener}}); err != nil {
-			ctx.Warnf("listen loop failed, retrying: %v", err)
+			ctx.Debugf("listen loop failed, retrying: %v", err)
 			return retry.RetryableError(err)
 		}
 		return nil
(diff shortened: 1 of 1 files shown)


postq/pg/notify.go
```diff
@@ -5,53 +5,98 @@ import (
...
+	// Group listeners by channel - multiple receivers can listen to same channel
+	channels := make(map[string][]chan<- string)
+	for _, listener := range listeners {
+		channels[listener.Channel] = append(channels[listener.Channel], listener.Receiver)
+	}
+
+	backoff := retry.WithMaxDuration(DBReconnectMaxDuration, retry.NewExponential(DBReconnectBackoffBaseDuration))
+	return retry.Do(ctx, backoff, func(retryCtx gocontext.Context) error {
+		if err := listenLoop(ctx, channels); err != nil {
+			ctx.Warnf("listen loop failed, retrying: %v", err)
```
moshloop:
```suggestion
			ctx.Debugf("listen loop failed, retrying: %v", err)
```

PR #1876, opened 2026-04-10
fix: only insert those configs that exist

yashmehrotra:
## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed an issue where configuration-change history entries could be recorded for configuration items that no longer exist or are marked inactive. The system now checks that a matching configuration item is present and active before creating a change record, preventing spurious history entries and improving data integrity.


moshloop:
@yashmehrotra I think we should record changes on deleted items, some changes may lag the actual deletion

